PR #37, opened 2026-04-21
feat: GitFS — mount any git repo as a read-only VFS backend

ivarvong:
## Summary

Adds `JustBash.FS.GitFS` — a read-only VFS backend backed by [ivarvong/exgit](https://github.com/ivarvong/exgit) (pure-Elixir git over smart HTTP v2, no `git` binary, no libgit2). Mount any public or private git repo at a path and use it from the shell.

- **Processless** — plain struct, no Agent, no ETS, no side effects. Fits the pure-value threading model of the rest of the codebase.
- **Fast mount (~300ms)** — single round trip via `filter: {:blob, :none}` partial clone. Refs, commits, and trees come back in one pack; blobs are fetched on demand.
- **Read-only** — all seven mutation callbacks return `:erofs` (defense in depth).
- **Optional dep** — `just_bash` compiles and runs without `exgit`. `GitFS.new/1` raises a clear error at runtime if the lib is absent.
- **Credentials** — any `Exgit.Credentials` value works directly (GitHub PAT, GitLab, Gitea, Bitbucket, basic, bearer).

## Performance

Benchmarks against `github.com/elixir-ai-tools/just_bash` (~110 files, 1.4 MB):

| Operation | Time | Notes |
|---|---|---|
| `GitFS.new/1` | **~300 ms** | 1 round trip: refs + commits + trees (single pack) |
| `GitFS.materialize/1` | **~500 ms** | 1 round trip: all reachable blobs |
| `ls` post-new | **<1 ms** | in-memory |
| `stat` post-new | **<1 ms** | in-memory |
| `cat` post-new | **~100 ms** | 1 round trip per blob (unless materialized) |
| `cat` post-materialize | **<1 ms** | in-memory |
| `grep -r` post-materialize | **~55 ms** | full recursive grep, no network |
| 100× `cat` post-materialize | **22 ms** | ~0.2 ms per call |

## Usage

```elixir
# Public repo, mounted at /repo
{:ok, repo_fs} = GitFS.new(url: "https://github.com/user/repo")
{:ok, fs} = FS.mount(FS.new(), "/repo", repo_fs)
bash = JustBash.new(fs: fs)

{r, _} = JustBash.exec(bash, "ls /repo")
{r, _} = JustBash.exec(bash, "cat /repo/README.md")
{r, _} = JustBash.exec(bash, "grep -r defmodule /repo/lib")

# Or mount as root
{:ok, repo_fs} = GitFS.new(url: "https://github.com/user/repo")
bash = JustBash.new(fs: FS.new(root: repo_fs))
```

### Grep-heavy workloads

```elixir
# One extra round trip up front; every subsequent read is in-memory
{:ok, state} = GitFS.new(url: "...")
{:ok, state} = GitFS.materialize(state)
```

### Private repos

```elixir
{:ok, state} = GitFS.new(
  url: "https://github.com/org/private-repo",
  auth: Exgit.Credentials.GitHub.auth("ghp_...")
)
```

### Cross-session caching

`%GitFS{}` is a pure value — hold it in app state, a GenServer, `:persistent_term`, or Cachex and share it across sessions. All sessions see the same object store with no locking:

```elixir
# Build once at app startup
{:ok, repo_fs} = GitFS.new(url: "...")

# All 1000 sessions share one clone
for _ <- 1..1000 do
  bash = JustBash.new(fs: FS.new(root: repo_fs))
  JustBash.exec(bash, "grep -r foo /")
end
```

## Network failures

`new/1` and `materialize/1` return `{:error, reason}` on network failure (no `MatchError` crashes):

- DNS failure → `{:error, %Req.TransportError{reason: :nxdomain}}` in ~100 ms
- TCP refused → `{:error, %Req.TransportError{reason: :econnrefused}}` in ~30 ms
- HTTP 4xx/5xx → `{:error, {:http_error, status, body}}` in ~180 ms
- Auth denied → `{:error, {:http_error, 401 | 403, _}}`

Connect timeout 10s, receive timeout 5min (configurable via `Exgit.Transport.HTTP`). Post-materialize, the struct is **fully offline-capable** — if GitHub goes down mid-session, everything still works.

## API

- `GitFS.new/1` → `{:ok, t} | {:error, term}` — clone a repo (partial by default, `eager: true` for full)
- `GitFS.new!/1` — raise-on-failure variant
- `GitFS.materialize/1` → `{:ok, t} | {:error, term}` — pull all reachable blobs into memory
- `GitFS.materialize!/1` — raise-on-failure variant

## Also in this PR

- `FS.new(root: backend_state)` now takes a plain struct (previously required `{module, state}` tuple — an inconsistency left over from before we dropped the explicit module arg in `FS.mount/3`)

## Tests

Two suites tagged `:live` (excluded from normal CI, require network + exgit dep):

- `test/just_bash/fs/git_fs_test.exs` — backend-level tests against `ivarvong/exgit`
- `test/just_bash/fs/git_fs_agent_test.exs` — agent-style exploration of `elixir-ai-tools/just_bash` (ls / cat / grep / wc / pipelines / stat). 17 tests, completes in **~1.5 s** after `materialize/1` in `setup_all`.

Run with:
```
mix test test/just_bash/fs/git_fs_agent_test.exs --include live
```

### Telemetry tracing

Set `EXGIT_TRACE=1` to print every exgit span event as it happens:

```
EXGIT_TRACE=1 mix test test/just_bash/fs/git_fs_agent_test.exs --include live
```

Sample output showing the exact network and in-memory operations:

```
[transport.ls_refs]  69 ms  url=.../just_bash ref_count=27 has_head=true
[transport.fetch]   158 ms  result_bytes=139_151 wants_count=1
[pack.parse]         12 ms  byte_size=139_151 object_count=677
[transport.fetch]   500 ms  result_bytes=2_630_460                # materialize
[pack.parse]         49 ms  byte_size=2_630_460 object_count=889
[fs.ls]            0.14 ms  path=lib/just_bash/commands entry_count=92
[fs.read_path]     0.10 ms  path=lib/just_bash/commands/echo.ex
```

## Dep

```elixir
{:exgit, github: "ivarvong/exgit", branch: "main", optional: true}
```

Will swap to a hex release before we tag.

## Follow-ups (explicitly deferred)

- `:path` option for persistent on-disk clones (needs proper open-or-clone logic, not just `Exgit.clone/2` which would duplicate work on restart)
- Symlink support (git has them at mode `120000`; we currently return `:einval` from `readlink` and `is_symbolic_link: false` from stat)
- Telemetry on error-collapse in `map_error/1` (operators will want to distinguish `:eio` causes)
- Add `:enotsup` to the `FS.Backend` contract's error atom list (LFS pointers)


mix.exs
```diff
@@ -52,7 +52,8 @@ defmodule JustBash.MixProject do
       {:dialyxir, "~> 1.4", only: [:dev, :test], runtime: false},
       {:credo, "~> 1.7", only: [:dev, :test], runtime: false},
       {:stream_data, "~> 1.0", only: [:dev, :test]},
-      {:benchee, "~> 1.0", only: :dev}
+      {:benchee, "~> 1.0", only: :dev},
+      {:exgit, github: "ivarvong/exgit", branch: "main", optional: true}
```
cjbell, line 56: We should publish this package

PR #39, opened 2026-06-10
feat: add JustBash.CLI for namespaced subcommand tools

### LINKED ISSUE #38: Feature Request: First-class CLI subcommand support for custom commands (routing, arg parsing, and help generation)
## Summary

Custom commands (`JustBash.Commands.Command`) are a flat, top-level namespace: one name → one `execute(bash, args, stdin)` that receives raw string args. There is no first-class way to build a namespaced CLI like:

```bash
acme pr review --report 1234
acme product list
```

This namespaced "tool family" style is extremely valuable for the library's stated primary audience — AI agents. A single `acme` entry point with discoverable, documented subcommands aids the agent far more than a pile of top-level commands (`acme_pr_review`, `acme_product_list`, …), and far more than each host application hand-rolling its own router.

I'm proposing a `JustBash.CLI` layer that handles **subcommand routing**, **typed argument parsing**, and **automatic help/documentation generation**, building on the existing `Command` behaviour so nothing about dispatch needs to change.

## Why this belongs in just_bash rather than userland

Today every consumer building a multi-command tool must write the same boilerplate. The repo's own eval suite already demonstrates this: [`eval/commands/kv.ex`](https://github.com/elixir-ai-tools/just_bash/blob/main/eval/commands/kv.ex) hand-rolls:

- a `case args do ["set", key | rest] -> ... ["get", key] -> ...` router,
- manual `--help` / `-h` / `[]` handling,
- a help string **duplicated verbatim** between `@moduledoc` and a private `help/1` function,
- ad-hoc `"kv: unknown subcommand '#{subcmd}'"` error formatting.

Every host app repeats this, each with slightly different conventions (help format, exit codes, error wording, `-h` vs `--help`, behavior on missing args). That inconsistency directly hurts agents, which do best with uniform, predictable tool UX.

Meanwhile the building blocks already exist inside the library but aren't exposed for this purpose:

- `JustBash.Commands.ArgParser` is a nice declarative flag parser (short/long forms, `:boolean`/`:string`/`:integer`/`:accumulator` types, defaults, transforms) — but it's an internal helper with no subcommand concept, no `required:` option, and no help-text generation.
- `JustBash.FlagParser` covers the simpler combined-short-flag style.
- The `Command` behaviour and the v0.3.0 `:context` option already give host commands everything they need at execution time.

The missing piece is the layer that ties these together.

### 1. A declarative CLI definition

A `use JustBash.CLI` macro (or an equivalent data-driven API, see "Alternatives" below) that compiles a command tree:

```elixir
defmodule Acme.CLI do
  use JustBash.CLI,
    name: "acme",
    description: "Acme operations toolkit"

  group "pr", about: "Pull request management" do
    command "review", about: "Review a pull request" do
      flag :report, type: :integer, required: true,
        doc: "ID of the report to review"
      flag :format, type: :string, default: "text",
        values: ["text", "json"], doc: "Output format"
      flag :verbose, short: "-v", type: :boolean

      run fn bash, opts, _positional, _stdin ->
        report = MyApp.PullRequests.fetch!(opts.report, bash.context.user)
        {JustBash.Commands.Command.ok(render(report, opts.format)), bash}
      end
    end
  end

  group "product", about: "Product lifecycle" do
    command "list", about: "List products" do
      run &MyApp.Products.list_cmd/4
    end

    command "show", about: "Show a product" do
      arg :id, required: true, doc: "Product ID"
      run &MyApp.Products.show_cmd/4
    end
  end
end

bash = JustBash.new(commands: %{"acme" => Acme.CLI})
```

The macro generates `names/0` and `execute/3`, so the module **is** a regular `JustBash.Commands.Command` — zero changes required to the executor, `normalize_commands!/1` validation, protected-builtin checks, or dispatch order (functions > custom > builtins). Arbitrary nesting depth should be supported (`group` inside `group`).

### 2. Routing and argument parsing

`execute/3` walks the tree on the leading positional tokens, then parses the remainder against the matched command's flag/arg spec (this can reuse and extend `ArgParser`):

- typed values (`:boolean`, `:string`, `:integer`, `:float`, `:accumulator`), defaults, `required:`, optional `values:` enums
- positional `arg` declarations with arity checking
- `--` to stop flag parsing, `--flag=value` and `--flag value` forms
- the handler receives `(bash, opts_map, positional, stdin)` — and keeps full access to `bash.fs`, `bash.context`, etc., exactly like today's custom commands

### 3. Generated help and errors (the agent-facing payoff)

All of the following come for free, with one consistent format across every CLI built on the feature:

```text
$ acme --help
acme - Acme operations toolkit

Usage: acme <command> [args]

Commands:
  pr review      Review a pull request
  product list   List products
  product show   Show a product

Run 'acme <command> --help' for details.

$ acme pr review --help
Usage: acme pr review --report <int> [--format text|json] [-v]

  --report <int>     ID of the report to review (required)
  --format <string>  Output format: text, json (default: text)
  -v, --verbose

$ acme pr reviw --report 1
acme: unknown command 'pr reviw'. Did you mean 'pr review'?
Run 'acme --help' for available commands.   # exit code 2

$ acme pr review
acme pr review: missing required flag --report
Usage: acme pr review --report <int> [--format text|json] [-v]   # exit code 2
```

Self-correcting error messages with usage lines are disproportionately valuable here: an agent that typos a subcommand or omits a flag recovers in one turn instead of flailing.

### 4. Introspection for agent prompting

Because the spec is declarative, expose it:

```elixir
JustBash.CLI.describe(Acme.CLI)
#=> %{name: "acme", description: "...", commands: [%{path: ["pr", "review"], flags: [...], ...}]}

JustBash.CLI.render_docs(Acme.CLI, format: :markdown)
```

This lets hosts generate the tool documentation that goes into the agent's system prompt **from the same source of truth** as the runtime behavior — eliminating the doc/implementation drift that the `kv.ex` example exhibits (its `@moduledoc` and `help/1` text are maintained by hand in parallel).

### 5. Shell-environment integration

Small touches that make the CLI feel native inside the sandbox:

- `type acme` / `command -v acme` — currently a custom command renders as the unhelpful `"acme is acme"` (`lib/just_bash/commands/type.ex`); CLI modules could report `"acme is a CLI (12 subcommands)"` or similar
- `help` integration if a `help` builtin is ever added
- **Telemetry**: `Telemetry.command_span/3` currently records only `command: "acme"` with raw args. The router should add the resolved subcommand path (e.g. `subcommand: ["pr", "review"]`) to span metadata so observability doesn't collapse an entire CLI into one bucket

## Alternatives considered

1. **Status quo (manual routing in `execute/3`)** — works, but every consumer rebuilds routing, parsing, help, and error UX; inconsistent across hosts; help text drifts from implementation; deep nesting gets ugly fast.
2. **Register each subcommand as a separate top-level command** (`"acme-pr-review" => ...`) — pollutes the global namespace and defeats the point: namespacing is what aids the agent.
3. **Shell functions in a sourced script** — can mimic `acme() { case "$1" in ... esac; }`, but functions run inside the sandbox and cannot reach host Elixir code, so this can't replace custom commands.
4. **A purely data-driven API instead of macros** — `JustBash.CLI.new(name: "acme", commands: [...])` with plain maps/structs and capture-form handlers. Arguably this should exist *as the foundation*, with the macro DSL as optional sugar on top; runtime-constructed CLIs (e.g., generated from a host app's own registry) would only be possible via the data API.

## Scope / non-goals

- No changes to the dispatch pipeline, protected builtins, or the existing `Command` behaviour — this is purely an additive layer that compiles down to it.
- No claim of sandboxing: CLI handlers are host code with the same trust model and caveats as today's custom commands (worth restating in the docs for this feature).
- Tab completion is out of scope (no interactive TTY), though `describe/1` would give hosts everything needed to build it if just_bash ever grows a REPL.

## Suggested acceptance criteria

- [ ] `use JustBash.CLI` produces a valid `JustBash.Commands.Command` (passes `normalize_commands!/1`), registrable under its root name plus optional aliases
- [ ] Nested `group`/`command` routing with correct precedence; unknown subcommand → exit 2 + usage + "did you mean" suggestion
- [ ] Typed flags and positionals with `required:`, defaults, enums, accumulators; parse errors → exit 2 + usage line
- [ ] Auto-generated `--help`/`-h` at every level of the tree, derived solely from the spec
- [ ] `JustBash.CLI.describe/1` and `render_docs/2` (`:text`, `:markdown`)
- [ ] Subcommand path included in `[:just_bash, :command]` telemetry metadata
- [ ] `type`/`command -v` aware of CLI modules
- [ ] Handlers retain full `bash` struct access (fs, env, context) and the existing crash-isolation behavior of `execute_custom_command/5`
- [ ] Docs: README section + a worked example (porting `eval/commands/kv.ex` to the new API would be a perfect before/after showcase)

## Environment

- just_bash v0.3.0, observations from `main` (`lib/just_bash.ex`, `lib/just_bash/commands/{command,arg_parser,type}.ex`, `lib/just_bash/interpreter/executor.ex`, `lib/just_bash/telemetry.ex`, `eval/commands/kv.ex`)

Happy to discuss the API shape — and I'd be glad to contribute a PR starting with the data-driven core + router, with the macro DSL as a follow-up.

davydog187:
Closes #38.

## Summary

Adds **`JustBash.CLI`** — a declarative, data-driven subcommand layer for building namespaced tools like `acme pr review --report 1234`. It compiles down to a value in the existing `commands` map, so nothing about dispatch, protected builtins, or the `Command` behaviour changes.

A CLI is plain data: a `%JustBash.CLI{}` tree of `JustBash.CLI.Command` nodes, built with `CLI.new/2` and `CLI.command/2`, run by a single engine (`CLI.run/4`).

```elixir
alias JustBash.CLI
alias JustBash.Commands.Command

cli =
  CLI.new("acme", doc: "Acme operations toolkit", commands: [
    CLI.command("pr", doc: "Pull request management", commands: [
      CLI.command("review",
        doc: "Review a pull request",
        flags: [
          report:  [type: :integer, required: true, doc: "ID of the report to review"],
          format:  [type: :string, default: "text", values: ~w(text json), doc: "Output format"],
          verbose: [type: :boolean, short: "-v"]
        ],
        run: &Acme.PR.review/1)
    ])
  ])

bash = JustBash.new(commands: %{"acme" => cli})
```

## Design notes

This is the issue's **Alternative #4** (data-driven core), taken further: the struct is the single source of truth and there is **no block DSL**. The library is ~86 plain `@behaviour` command modules with exactly one macro in `lib/` (the `~b` sigil), so a `group do … command do … end end` DSL would be the only construct of its kind. Instead:

- `group`/`command` are unified into one nestable `command` (a node with children is a group; a node with a `run` is a leaf) — matching git/cargo/kubectl.
- `doc:` is used everywhere (not `about:` for nodes and `doc:` for flags).
- Handlers take a single `%JustBash.CLI.Invocation{}` (`flags`, `args`, `bash`, `stdin`, `path`) and return `{result, bash}` — the same contract as a custom command, keeping full access to `bash.fs`/`bash.context`.
- An optional `use JustBash.CLI` + `spec/0` lets a CLI live as a module (GenServer-style wiring, not a DSL). Registering the struct directly works too; both run on the same engine.

## What's included

- **Routing engine** `CLI.run/4`: nested routing, typed flag parsing, positional arity (incl. variadic), `--`/`--flag=value` forms.
- **Parser extensions** to `ArgParser`: `:float`, `:required`, `:values` (enum) — all opt-in, existing callers unchanged and regression-tested.
- **Help + errors**: auto `--help`/`-h` at every level; unknown subcommand → exit 2 + usage + `Did you mean …?` (`String.jaro_distance`); missing/invalid args → exit 2 + usage line.
- **Introspection**: `CLI.describe/1` (nested map) and `CLI.render_docs/2` (`:text`/`:markdown`) — same source of truth as runtime, so agent-prompt docs can't drift.
- **Shell integration**: `type` / `command -V` report `"acme is a CLI tool (N commands)"`.
- **Telemetry**: `[:just_bash, :command]` stop metadata carries the resolved `subcommand: ["pr","review"]` path (transient result key, read into span metadata then stripped — no contract change).
- **Showcase**: `eval/commands/kv.ex` ported to `eval/commands/kv_cli.ex` (before/after), plus a README section.

## Acceptance criteria (from #38)

- [x] `%CLI{}`/`use`-module registers as a valid command (passes `normalize_commands!`), with aliases
- [x] Nested routing + precedence; unknown subcommand → exit 2 + usage + did-you-mean
- [x] Typed flags/positionals with `required`, defaults, enums, accumulators; parse errors → exit 2 + usage
- [x] Auto-generated `--help`/`-h` at every level, from the spec
- [x] `describe/1` and `render_docs/2` (`:text`, `:markdown`)
- [x] Subcommand path in `[:just_bash, :command]` telemetry metadata
- [x] `type` / `command -v` aware of CLI modules
- [x] Handlers retain full `bash` access and crash isolation
- [x] Docs: README section + `kv.ex` before/after

## Also

Fixes a pre-existing flaky telemetry test (`telemetry_test.exs:103`) that matched a session `:stop` event without binding the session pid — a concurrent async test's event could land in its mailbox. Now binds the pid from `:start`, as its sibling test already does.

## Verification

All five gates pass: `mix compile --warnings-as-errors`, `mix format --check-formatted`, `mix credo --strict` (only the pre-existing intentional fixture finding), `mix test` (**3607 tests, 0 failures**), `mix dialyzer` (clean).

## Behavior change (review follow-up — dab60ef)

Hardening from code review. The one item worth release-note attention:

- **`ArgParser` `:integer`/`:float` parsing is now strict.** A value with trailing
  characters is rejected instead of silently truncated: `--max-time 30s` / `--report 42x`
  / `1_000` now error (`invalid integer value: …`) rather than parsing as `30`/`42`/`1`.
  This matches real CLI behavior (`curl --max-time 30s` errors). The only existing
  `:integer` consumer is `curl` (`--max-time`, `--connect-timeout`); the full suite passes,
  so nothing depended on the lenient behavior.

Other review fixes (no external behavior impact): reject `:flags`/`:args` on group nodes,
tag the resolved subcommand path on help/usage-error telemetry, consume a leading `--`
options terminator in routing, and raise a clear error on a malformed handler return.

### DIFF
--- a/lib/just_bash/cli.ex
+++ b/lib/just_bash/cli.ex
@@ -44,6 +44,40 @@ defmodule JustBash.CLI do
   `:float`, `:accumulator`), `:short`, `:long` (defaults to `--name`), `:default`,
   `:required`, `:values` (enum), `:transform`, and `:doc` (used in help output).
 
+  Flag specs are validated at build time (`command/2` raises `ArgumentError`):
+
+    * `--help`/`-h` are **reserved** — see "Reserved flags" below.
+    * `:required` and `:default` are mutually exclusive (a required flag errors when omitted,
+      so the default could never apply).
+    * a `:default` must be a member of `:values` when both are given (the enum check only
+      runs on flags the user actually provides, so an out-of-range default would slip past it).
+
+  `:values` is compared against the **coerced** value, so list members must match the flag's
+  `:type` — e.g. `type: :integer, values: [1, 2]` (integers, not `~w(1 2)`). The raw
+  `:values` list is also what `describe/1` and the help text surface to agents.
+
+  ## Reserved flags
+
+  `--help` and `-h` are reserved: the router intercepts them before a leaf ever parses, so a
+  leaf can request help in one turn. A flag spec may not claim either form (including a flag
+  named `:help`, whose derived long is `--help`) — `command/2` raises if it does. Because
+  interception happens first, `--help`/`-h` anywhere before a `--` terminator wins even when
+  it would otherwise be a flag's value (e.g. `acme pr review --format -h` shows help).
+
+  ## `--` handling
+
+  A leading `--` *before* a subcommand is consumed and routing continues (`acme -- pr review`
+  reaches `pr review`), so wrappers that prepend `--` still route. This is intentionally not
+  POSIX `--` semantics — `--` only acts as an options/help terminator once routing reaches a
+  leaf and hands the remaining tokens to the parser.
+
+  ## Positional arguments
+
+  Positionals are a flat list, so `command/2` rejects ambiguous shapes at build time: a
+  required positional may not follow an optional one, and a variadic must be last. A lone
+  `-` (the stdin convention) is treated as a flag by the router and is **not** supported as a
+  positional; pass it after `--` if a leaf needs it as a literal value.
+
   ## Trust model
 
   CLI handlers are ordinary host Elixir code with the same trust model and crash
@@ -108,7 +142,10 @@ defmodule JustBash.CLI do
       @behaviour JustBash.CLI
 
       @impl JustBash.Commands.Command
-      def names, do: [spec().name | spec().aliases]
+      def names do
+        s = spec()
+        [s.name | s.aliases]
+      end
 
       @impl JustBash.Commands.Command
       def execute(bash, args, stdin), do: JustBash.CLI.run(spec(), bash, args, stdin)
@@ -135,6 +172,9 @@ defmodule JustBash.CLI do
   @spec custom_command_description(term(), String.t()) :: String.t()
   def custom_command_description(value, name) do
     if cli?(value) do
+      # `describe/1` flattens to leaves only (groups are recursed into, not listed), so this
+      # counts runnable commands, not tree nodes. If `describe/1`'s shape ever changes, the
+      # "describes a CLI tool" tests in test/cli/shell_integration_test.exs pin this count.
       count = value |> describe() |> Map.fetch!(:commands) |> length()
       "#{name} is a CLI tool (#{count} #{pluralize(count, "command", "commands")})"
     else
@@ -361,8 +401,12 @@ defmodule JustBash.CLI do
   # %CLI{} or a %Command{} group; both expose `name`/`doc`/`commands`.
   defp route(group, [], path), do: {:no_subcommand, group, path, []}
 
-  # A `--` options terminator before a subcommand is consumed; routing continues with the
-  # rest, so `acme -- pr review` reaches `pr review` instead of erroring as a stray flag.
+  # A leading `--` before a subcommand is consumed; routing continues with the rest, so
+  # `acme -- pr review` reaches `pr review` instead of erroring as a stray flag. This is
+  # deliberately NOT POSIX `--` semantics (which would make everything after it positional):
+  # it lets wrappers that prepend `--` still route to subcommands. `--` only acts as an
+  # options/help terminator once routing reaches a leaf and hands the rest to the parser.
+  # See the "`--` handling" note in the moduledoc.
   defp route(group, ["--" | rest], path), do: route(group, rest, path)
 
   defp route(group, ["-" <> _ | _] = remaining, path),
@@ -534,7 +578,9 @@ defmodule JustBash.CLI do
               "command #{inspect(name)} flag #{inspect(flag_name)} spec must be a keyword list, got: #{inspect(spec)}"
       end
 
-      {flag_name, Keyword.put_new(spec, :long, default_long(flag_name))}
+      spec = Keyword.put_new(spec, :long, default_long(flag_name))
+      validate_flag_spec!(name, flag_name, spec)
+      {flag_name, spec}
     end)
   end
 
@@ -543,21 +589,71 @@ defmodule JustBash.CLI do
           "command #{inspect(name)} :flags must be a keyword list of flag specs, got: #{inspect(flags)}"
   end
 
+  # Build-time guards that can't drift into runtime surprises:
+  #   * `--help`/`-h` are intercepted by the router before any leaf parses, so a flag that
+  #     claims them could never receive its value (see the "reserved flags" note in the
+  #     moduledoc).
+  #   * `:required` + `:default` is contradictory — the default can never apply, since a
+  #     required flag errors when omitted.
+  #   * a `:default` outside `:values` would silently bypass the enum check, which only
+  #     runs on provided flags.
+  defp validate_flag_spec!(name, flag_name, spec) do
+    cond do
+      spec[:long] == "--help" or spec[:short] == "-h" ->
+        reserved = if spec[:long] == "--help", do: "--help", else: "-h"
+
+        raise ArgumentError,
+              "command #{inspect(name)} flag #{inspect(flag_name)}: #{reserved} is reserved for help and cannot be used as a flag"
+
+      spec[:required] && Keyword.has_key?(spec, :default) ->
+        raise ArgumentError,
+              "command #{inspect(name)} flag #{inspect(flag_name)} cannot be both :required and have a :default"
+
+      true ->
+        validate_default_in_values!(name, flag_name, spec)
+    end
+  end
+
+  defp validate_default_in_values!(name, flag_name, spec) do
+    with {:ok, default} <- Keyword.fetch(spec, :default),
+         values when is_list(values) <- spec[:values],
+         false <- default in values do
+      raise ArgumentError,
+            "command #{inspect(name)} flag #{inspect(flag_name)}: :default #{inspect(default)} " <>
+              "is not one of :values #{inspect(values)}"
+    else
+      _ -> :ok
+    end
+  end
+
   defp default_long(flag_name) do
     "--" <> String.replace(Atom.to_string(flag_name), "_", "-")
   end
 
   defp validate_args!(name, args) when is_list(args) do
-    {normalized, _seen_variadic} =
-      Enum.map_reduce(args, false, fn arg, variadic_seen? ->
+    {normalized, _acc} =
+      Enum.map_reduce(args, %{variadic?: false, optional?: false}, fn arg, acc ->
         spec = normalize_arg!(name, arg)
 
-        if variadic_seen? do
+        if acc.variadic? do
           raise ArgumentError,
                 "command #{inspect(name)}: a variadic positional argument must be last"
         end
 
-        {spec, spec.variadic}
+        # A required positional after an optional one is ambiguous: the handler receives a
+        # flat list and can't tell which slot a single value bound to. Reject it at build
+        # time, mirroring the variadic-must-be-last rule.
+        if spec.required and acc.optional? do
+          raise ArgumentError,
+                "command #{inspect(name)}: required positional argument #{inspect(spec.name)} " <>
+                  "cannot follow an optional one"
+        end
+
+        {spec,
+         %{
+           variadic?: acc.variadic? or spec.variadic,
+           optional?: acc.optional? or not spec.required
+         }}
       end)
 
     normalized
--- a/lib/just_bash/cli.ex
+++ b/lib/just_bash/cli.ex
@@ -221,6 +221,7 @@ defmodule JustBash.CLI do
     args = Keyword.get(opts, :args, [])
 
     validate_node_kind!(name, commands, run)
+    validate_leaf_only!(name, commands, flags, args)
     validate_commands!(commands)
     validate_unique_names!(commands, name)
     flags = normalize_flags!(name, flags)
@@ -248,23 +249,27 @@ defmodule JustBash.CLI do
   """
   @spec run(t(), JustBash.t(), [String.t()], String.t()) :: {map(), JustBash.t()}
   def run(%__MODULE__{} = cli, bash, args, stdin) do
+    # Every routed result carries its resolved subcommand path (the valid prefix, even
+    # for help/usage errors) so command telemetry can attribute the bucket; the executor
+    # reads it into span metadata and strips it before the result reaches the shell.
     case route(cli, args, []) do
       {:leaf, leaf, path, rest} ->
         if wants_help?(rest) do
-          {help_result(Help.leaf_help(cli, path, leaf)), bash}
+          {tag_subcommand(help_result(Help.leaf_help(cli, path, leaf)), path), bash}
         else
           dispatch_leaf(cli, leaf, path, rest, bash, stdin)
         end
 
       {:no_subcommand, group, path, remaining} ->
         if wants_help?(remaining) do
-          {help_result(Help.group_help(cli, path, group)), bash}
+          {tag_subcommand(help_result(Help.group_help(cli, path, group)), path), bash}
         else
-          {usage_error(Help.missing_subcommand(cli, path, group)), bash}
+          {tag_subcommand(usage_error(Help.missing_subcommand(cli, path, group)), path), bash}
         end
 
       {:unknown_subcommand, group, path, token} ->
-        {usage_error(Help.unknown_subcommand(cli, path, group, token)), bash}
+        {tag_subcommand(usage_error(Help.unknown_subcommand(cli, path, group, token)), path),
+         bash}
     end
   end
 
@@ -356,6 +361,10 @@ defmodule JustBash.CLI do
   # %CLI{} or a %Command{} group; both expose `name`/`doc`/`commands`.
   defp route(group, [], path), do: {:no_subcommand, group, path, []}
 
+  # A `--` options terminator before a subcommand is consumed; routing continues with the
+  # rest, so `acme -- pr review` reaches `pr review` instead of erroring as a stray flag.
+  defp route(group, ["--" | rest], path), do: route(group, rest, path)
+
   defp route(group, ["-" <> _ | _] = remaining, path),
     do: {:no_subcommand, group, path, remaining}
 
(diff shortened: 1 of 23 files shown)


lib/just_bash/commands/arg_parser.ex
```diff
@@ -227,30 +261,57 @@ defmodule JustBash.Commands.ArgParser do
...
 
+  defp coerce_value(:string, value), do: {:ok, value}
+  defp coerce_value(:accumulator, value), do: {:ok, value}
+
+  defp coerce_value(:boolean, value), do: {:ok, value in ["true", "1", "yes"]}
+
+  defp coerce_value(:integer, value) do
+    case Integer.parse(value) do
+      {n, ""} -> {:ok, n}
+      {n, _} -> {:ok, n}
```
davydog187: These two clauses produce the identical result, so the `{n, ""}` clause is dead — `{n, _}` already subsumes it. The split reads as if it were *meant* to reject trailing garbage but doesn't: `Integer.parse("42x")` → `{42, "x"}` → `{:ok, 42}`, and `Integer.parse("1_000")` → `{1, "_000"}` → `{:ok, 1}`.

This matches the pre-existing behavior (the old code also did `{n, _} -> ...`), so it's not a regression. But for a CLI layer whose stated goal is letting an agent recover from a typo in one turn, silently truncating `--report 42x` to `42` works against that. Consider making it strict:

```elixir
defp coerce_value(:integer, value) do
  case Integer.parse(value) do
    {n, ""} -> {:ok, n}
    _ -> {:error, "invalid integer value: #{value}\n"}
  end
end
```

If you'd rather not change behavior for existing `ArgParser` callers, at minimum collapse to a single `{n, _}` clause so the redundancy doesn't read as an unfinished intent.

davydog187 (commented):
Reviewed the CLI subcommand layer end-to-end. This is excellent work — the data-driven design (no block DSL, struct as single source of truth) is the right call for this codebase, the moduledocs are exemplary, and test coverage is thorough across routing, help, introspection, the `use` macro, shell integration, and telemetry. The `ArgParser` extensions are cleanly opt-in and regression-tested, and folding the subcommand path through telemetry via a stripped transient key is a tidy trick.

Findings are all minor — mostly edge cases in numeric coercion and a couple of validation gaps. None block merge. Details inline.

One meta note: PR body says `Closes #38` but #38 doesn't exist in this repo (this is PR #39). Worth fixing the reference so it doesn't dangle.

lib/just_bash/commands/arg_parser.ex
```diff
@@ -227,30 +261,57 @@ defmodule JustBash.Commands.ArgParser do
...
+      :error -> {:error, "invalid integer value: #{value}\n"}
+    end
+  end
+
+  defp coerce_value(:float, value) do
+    case Float.parse(value) do
+      {f, _} -> {:ok, f}
+      :error -> {:error, "invalid float value: #{value}\n"}
+    end
+  end
```
davydog187, line 290: Same trailing-garbage tolerance as `:integer`: `Float.parse("1.5.6")` → `{1.5, ".6"}` → `{:ok, 1.5}`. Since `:float` is brand new in this PR (no existing callers to preserve), this is the cheapest place to be strict from day one — match on `{f, ""}` and error otherwise. Whatever you decide here, `:integer` and `:float` should behave consistently.

lib/just_bash/cli.ex
```diff
@@ -0,0 +1,559 @@
...
+  @spec command(String.t(), keyword()) :: Command.t()
+  def command(name, opts \\ []) when is_binary(name) do
+    validate_name!(name, "command")
+
+    commands = Keyword.get(opts, :commands, [])
+    run = Keyword.get(opts, :run)
+    flags = Keyword.get(opts, :flags, [])
+    args = Keyword.get(opts, :args, [])
+
+    validate_node_kind!(name, commands, run)
```
davydog187, line 263: Validation gap: `validate_node_kind!` enforces group-xor-leaf, but nothing rejects `:flags`/`:args` on a **group** node. `CLI.command("pr", commands: [...], flags: [report: [...]])` is accepted, the flags are normalized and stored on the struct, and then silently ignored — routing never consults a group's flags/args. That's a quiet footgun for someone who expects a parent-level flag to apply. Consider raising when a group is given `:flags`/`:args`, or at least documenting that they're leaf-only and dropped on groups.

lib/just_bash/cli.ex
```diff
@@ -0,0 +1,559 @@
...
+        else
+          dispatch_leaf(cli, leaf, path, rest, bash, stdin)
+        end
+
+      {:no_subcommand, group, path, remaining} ->
+        if wants_help?(remaining) do
+          {help_result(Help.group_help(cli, path, group)), bash}
+        else
+          {usage_error(Help.missing_subcommand(cli, path, group)), bash}
+        end
```
davydog187, line 308: Telemetry observability gap: only `dispatch_leaf` tags the result with `__subcommand__`, so the `:no_subcommand`, `:unknown_subcommand`, and `--help` paths emit no `subcommand` metadata. Given the PR's framing — "observability doesn't collapse an entire CLI into a single `command` bucket" — a flood of `acme pr <typo>` or `acme pr --help` calls would be exactly the bucket you'd want to see, and they're invisible here. Not a correctness issue; flagging in case partial subcommand attribution (e.g. tag the routed `path` even on unknown/no-subcommand) is desirable.

lib/just_bash/cli.ex
```diff
@@ -0,0 +1,559 @@
...
+  end
+
+  # --- routing ---
+
+  # Walks the tree consuming leading positional tokens. `group` is either the root
+  # %CLI{} or a %Command{} group; both expose `name`/`doc`/`commands`.
+  defp route(group, [], path), do: {:no_subcommand, group, path, []}
+
+  defp route(group, ["-" <> _ | _] = remaining, path),
+    do: {:no_subcommand, group, path, remaining}
```
davydog187, line 413: Minor: this treats any `-`-prefixed leading token as "stop routing," which means a `--` options-terminator before a subcommand (`acme -- pr review`) is captured as a flag-ish remainder and resolves to a missing/invalid-subcommand error rather than routing to `pr review`. Bash users may reach for `--` out of habit. Low priority — just noting the interaction with `wants_help?`, which also stops at `--`.

lib/just_bash/cli.ex
```diff
@@ -0,0 +1,559 @@
...
+        bash: bash,
+        flags: flags,
+        args: positional,
+        stdin: stdin,
+        path: path
+      }
+
+      case leaf.run.(invocation) do
+        {result, new_bash} -> {tag_subcommand(result, path), new_bash}
+        other -> other
```
davydog187: `other -> other` passes a malformed handler return (anything not a `{result, bash}` tuple) straight through to `execute_custom_command`, which will fail its strict shape match. It works out because the executor turns that into an error, but the failure surfaces far from the cause. Consider validating the tuple here so a buggy handler gets a clear, CLI-attributed message. Optional.

test/telemetry_test.exs
```diff
@@ -105,8 +105,13 @@ defmodule JustBash.TelemetryTest do
       script = "echo hello; echo world"
       {result, _bash} = JustBash.exec(bash, script)
 
+      # Bind this run's session pid from the start event so a concurrent async test's
+      # session :stop (delivered to us by the globally-attached handler) can't be matched
+      # here by mistake.
```
davydog187, line 110: Nice catch on the flaky test — binding the session pid from the `:start` event before asserting on `:stop` is the correct fix for a globally-attached handler racing with concurrent async tests. The inline comment explaining *why* is exactly right.

lib/just_bash/commands/arg_parser.ex
davydog187: Fixed in dab60ef — `:integer` now requires the whole value to parse (`{n, ""}`), so `42x` and `1_000` error instead of truncating to 42/1. Failing-first tests added in arg_parser_test.exs.

lib/just_bash/commands/arg_parser.ex
davydog187: Fixed in dab60ef — `:float` made strict to match `:integer`; `1.5.6` now errors. Test added.

lib/just_bash/cli.ex
davydog187: Fixed in dab60ef — `command/2` now raises via `validate_leaf_only!/4` when a group node is given `:flags`/`:args`, instead of silently dropping them. Tests in builder_test.exs.

lib/just_bash/cli.ex
davydog187: Fixed in dab60ef — `run/4` now tags the resolved path on the help, no-subcommand, and unknown-subcommand results too (empty path for the bare root), so those buckets show up in telemetry. Tests added in shell_integration_test.exs.

lib/just_bash/cli.ex
davydog187: Fixed in dab60ef — `route/3` consumes a leading `--` and continues routing, so `acme -- pr review` reaches the leaf instead of erroring as a stray flag. Test in routing_test.exs.

lib/just_bash/cli.ex
davydog187: Fixed in dab60ef — a non-`{result, bash}` return now raises a clear, CLI-attributed `ArgumentError` (caught by the executor → exit 1) instead of the generic "invalid result". Test in routing_test.exs.

davydog187:
> This is excellent work

dave-medal-dave

sorentwo:
Indeed—nicely done—Lave Ducia

5694de87-2c40-4d2e-b016-ff221a781380

ivarvong (approved):

PR #44, opened 2026-07-01
feat!: replace the filesystem layer with the vfs library

ivarvong:
Swaps JustBash's filesystem substrate for [vfs 0.1.0](https://hex.pm/packages/vfs). `bash.fs` is now a `%VFS{}` mount table with `JustBash.FS.Memory` — a `VFS.Mountable` port of the old `InMemoryFs` that keeps symlinks, hard links, modes, and mtimes — mounted at `/`.

## What this PR does

- **`JustBash.FS.Memory`** implements the ten-op `VFS.Mountable` protocol over the old entry model (files/dirs/symlinks with mode + mtime). Symlink resolution (with loop detection) happens inside the backend; `stat` follows links, `write_file` honors `mode:`/`mtime:` opts.
- **`JustBash.FS.POSIX`** is a secondary protocol (per the vfs SPEC's escape-hatch pattern) carrying the ops vfs deliberately cut: `lstat`, `symlink`, `readlink`, `link`, `chmod`, `append_file`. The `VFS` impl routes through the mount table exactly like the core ops (mount-prefix stripping, state threading, error-path rewriting); the `Any` fallback degrades gracefully — `lstat` ≡ `stat`, `chmod` is a validated no-op, `append_file` composes read+write, `symlink`/`link` return `:enotsup`. So `ln -s` on an exgit mount will fail with "Operation not supported" instead of crashing.
- **`JustBash.FS`** is the thin facade: core ops delegate to `VFS` (vfs-shaped returns — reads return `{:ok, payload, fs}` and every consumer threads the updated fs forward, so lazy backends keep their read caches), path helpers wrap `VFS.Path`, and `cp`/`mv` become compositions over the primitives — which makes them work **across mounts**.
- **`JustBash.mount/3` / `JustBash.umount/2`**: any `VFS.Mountable` mounts into a running environment and every bash command sees it — `cat`, globs, redirections, conditionals, `cd`. The exgit follow-up becomes `JustBash.mount(bash, "/repo", repo)`.
- All ~50 command modules, the interpreter core (redirections, `[[ ]]` file tests, glob expansion), and the eval harness migrated to `%VFS.Stat{type: atom}` / `%VFS.Error{kind: atom}` shapes. `JustBash.FS.strerror/1` centralizes kind→message text.
- New tests: `fs_test.exs` (ports the old backend suite to the facade + mount-table routing, degradation, cross-mount `:exdev` hard links) and `mount_integration_test.exs` (bash commands end-to-end over a foreign `VFS.Memory` mount).

## Worked examples

- `test/showcase_test.exs` — runnable guided tour (value semantics, FS-threading custom commands, mounts, degradation): `mix test test/showcase_test.exs --trace`
- `UPGRADING.md` — full 0.3 → 0.4 mapping for host-side code that touches `bash.fs` (in hexdocs extras and the package tarball)
- README — custom-command example written against the new `JustBash.FS` API, plus a "Virtual Filesystem and Mounts" section

## Breaking changes (pre-1.0)

Breaking **only for host-side Elixir code that touches `bash.fs`** (custom commands, test helpers): `JustBash.Fs` / `JustBash.Fs.InMemoryFs` become `JustBash.FS` / `JustBash.FS.Memory` (the two spellings cannot coexist as deprecated aliases — same beam filename on case-insensitive filesystems); reads return `{:ok, payload, fs}`, errors are `%VFS.Error{}`, stat is `%VFS.Stat{}`, `exists?` returns `{boolean, fs}`; `mkdir recursive:` → `parents:`, `rm force:` removed. `JustBash.new/exec/exec_file`, the sigil, `:context`, and `JustBash.CLI` are unchanged.

Bash-level behavior is unchanged except seven verified deltas, each enumerated with its 0.3 behavior in UPGRADING.md ("Exact behavioral changes at the bash level") — headline items: appends (`>>`, `tee -a`) now follow symlinks to the target per POSIX `O_APPEND` (0.3's `tee -a link` silently **dropped the target's content**; fixed here test-first), `>>` preserves the target's mode, `cat` on a symlink loop errors instead of raising out of `exec/2`, `-nt`/`-ot` compare mtimes chronologically, and a few error strings use strerror text instead of raw atoms.

## Auditing implementer code

Most legacy shapes compile cleanly and misbehave silently (a stale `{:ok, content}` match falls through to the error clause; `FS.exists?/2` in an `if` is an always-truthy tuple). Two defenses ship in this PR:

- **`mix just_bash.audit [paths]`** — AST scanner covering all seven legacy shapes (module refs, stale 2-tuple matches, atom-error clauses, truthy `exists?`, legacy opts, `stat.is_file`-style fields, `bash.fs.data`), alias- and pipe-aware, `file:line` output with a fix hint, non-zero exit for CI. Runs clean on `lib/` + `eval/`; deliberately catches the two legacy-shape guard tests when pointed at `test/`.
- **Runtime guards** — the two shapes that would be silently *ignored* (`mkdir recursive:`, `rm force:`) now raise `ArgumentError` pointing at UPGRADING.md.

## vfs 0.2 feedback from this consumer

- `VFS.Error`'s kind union is missing `:enotempty` (POSIX `rmdir(2)`); `rm` of a non-empty directory needs it. Constructed out-of-contract for now with two dialyzer skips annotated to be removed when vfs adds it.
- The POSIX routing defimpl uses `VFS.__resolve__/2` / `VFS.__put_mount__/3`; a blessed public API for secondary protocols that need mount resolution would remove the dunder dependency.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016JoScNjjVR8xvzkAbEpb9M


lib/just_bash/migration_audit.ex
```diff
@@ -0,0 +1,334 @@
+defmodule JustBash.MigrationAudit do
```
davydog187: do you intend to this this as part of the public API?

davydog187: seems cool but maybe only pulled in if you are using the mix task?

lib/just_bash/fs/memory.ex
```diff
@@ -0,0 +1,503 @@
...
+  Parent directories are created automatically.
+
+  ## Examples
+
+      iex> fs = JustBash.FS.Memory.new()
+      iex> fs = JustBash.FS.Memory.new(%{"/home/user/file.txt" => "hello"})
+      iex> fs = JustBash.FS.Memory.new(%{"/bin/script" => %{content: "#!/bin/bash", mode: 0o755}})
+  """
+  @spec new(map()) :: t()
+  def new(initial_files \\ %{}) do
```
davydog187, line 88: I might be missing something, but I would have expected the functions in this module to be delegating to VFS

davydog187: Is this a legacy module? It says its the JustBash in memory backend for VFS. Doesn't VFS implement that already?

lib/just_bash/migration_audit.ex
ivarvong: Good call — it's migration-window tooling, not durable API. Moved in 6678360: the scanner is now `Mix.Tasks.JustBash.Audit.Scanner` with `@moduledoc false` (implementation detail of the task), the rule table lives in the task's moduledoc (`mix help just_bash.audit`), and it's out of the hexdocs module list. The task moduledoc also now states it will be removed once the 0.3-era shapes are gone — same lifetime as the `mkdir recursive:`/`rm force:` runtime guards.

lib/just_bash/fs/memory.ex
ivarvong: Not legacy, and it can't delegate — this is the "add when there's a real consumer" case from the vfs SPEC, and just_bash is that consumer. vfs 0.1 deliberately cut lstat/symlink/readlink/link/chmod/append_file from the protocol, and `VFS.Memory` stores bare `path => binary` pairs with no entry metadata. Bash needs exactly what was cut (`ln -s`, `readlink`, `chmod`, `test -L`, `ls -l` modes, mtime-honoring writes), which requires a richer entry model — file/directory/symlink entries carrying mode + mtime, with link resolution — that can't be layered over a flat binary map. So this backend owns its storage, implements the ten `VFS.Mountable` ops against it, and carries the POSIX extras via the `JustBash.FS.POSIX` secondary protocol (which plain `VFS.Memory` mounts refuse gracefully). Rewrote the moduledoc in 6678360 to answer exactly this question up front — it clearly wasn't doing its job if it read as legacy.

PR #56, opened 2026-07-29
fix: writing through a regular file is ENOTDIR, not unreachable state

### LINKED ISSUE #53: 0.4: the store holds paths FS.walk/3 cannot reach; writing through a regular file should be ENOTDIR
On **main** (`56c74b8`, post-#44) a write through a path that is a regular file succeeds, stores the child, and then `FS.walk/3` never yields it. `read_file/2` on that same path succeeds. So the filesystem holds state that enumeration cannot reach.

This is a **regression from 0.3.0**, where the equivalent path was enumerable.

## Repro

```elixir
b = JustBash.new(files: %{"/m/j" => "parent\n"})
{r, b2} = JustBash.exec(b, "echo hi > /m/j/2026/a.md")

r.exit_code  #=> 0        real bash: exit 1, "bash: /m/j/2026/a.md: Not a directory"
r.stderr     #=> ""

JustBash.FS.read_file(b2.fs, "/m/j/2026/a.md")
#=> {:ok, "hi\n", _fs}                            — stored and readable

b2.fs |> JustBash.FS.walk("/m", include_dirs: true) |> Enum.map(&elem(&1, 0))
#=> ["/m", "/m/j"]                                 — child never yielded
```

`/m/j` remains a `:file` in the backing map while `/m/j/2026/a.md` sits beside it, so `walk` correctly declines to descend through a regular file — but the entry is still there to be read.

### 0.3.0 for comparison

```elixir
InMemoryFs.get_all_paths(b2.fs)
#=> ["/m", "/m/j", "/m/j/2026", "/m/j/2026/a.md"]  — enumerable
```

Same permissive write, but nothing was hidden. The 0.4 tree-shaped traversal is what turns it into unreachable state.

## The root cause: writes don't check ENOTDIR

The unreachable entry can only exist because the write was allowed. Real bash refuses all three of these; just_bash allows all three, in **both 0.3.0 and main**:

| command (`/m/j` is a regular file) | real bash | just_bash |
|---|---|---|
| `echo hi > /m/j/a.md` | exit 1, `Not a directory` | exit 0 |
| `echo hi > /m/j/2026/a.md` | exit 1, `Not a directory` | exit 0 |
| `mkdir -p /m/j/2026` | exit 1, `mkdir: /m/j: Not a directory` | exit 0 |

Fixing this makes the unreachable state impossible by construction, which is why I'd treat it as one issue rather than two. But the enumeration inconsistency is arguably a bug on its own terms even if the permissive write were considered an acceptable simplification of a flat store — a readable path that cannot be enumerated has no POSIX analogue.

Relatedly, `JustBash.new/1` now *accepts* `%{"/m/j" => _, "/m/j/a.md" => _}`, where 0.3.0 raised a `MatchError` inside the mount. That's a nicer failure mode in isolation, but combined with the above it converts a loud crash into silence.

## Why it matters downstream

[dol](https://github.com/davydog187/dol) embeds the sandbox and persists `/memory` by walking each mount root after every command. With 0.4 the agent can run one exit-0 command and have the write silently not persisted **and** absent from the "these files couldn't be saved" report it shows the model — so the model builds on a file that isn't there. On 0.3 the same situation raised loudly at mount time, which is worse ergonomically but at least not silent data loss. That's what's currently blocking dol from taking 0.4 ([davydog187/dol#16](https://github.com/davydog187/dol/pull/16)).

Happy to send a PR for the ENOTDIR checks on `write_file` / `mkdir` paths if that's the direction you'd want.

davydog187:
Closes #53.

## Root cause

`JustBash.FS.Memory` stores entries in a flat `path => entry` map, and
`ensure_parent_dirs/2` walked up from the target creating missing
directories, stopping at the first ancestor that *existed* — without
checking that it was a directory. Nothing anywhere in the write path
enforced POSIX's rule that every non-final path component must be a
directory.

So with `/m/j` a regular file:

```elixir
b = JustBash.new(files: %{"/m/j" => "parent\n"})
{r, b2} = JustBash.exec(b, "echo hi > /m/j/2026/a.md")
r.exit_code                                        #=> 0   (bash: 1)
JustBash.FS.read_file(b2.fs, "/m/j/2026/a.md")     #=> {:ok, "hi\n", _}
b2.fs |> JustBash.FS.walk("/m", include_dirs: true) |> Enum.map(&elem(&1, 0))
#=> ["/m", "/m/j"]     — the entry is stored, readable, and unreachable
```

`read_file/2` hits the map key directly, while `readdir/2` and `walk/3`
have to route through `/m/j` — a file, not a directory — so they can never
yield the child. The filesystem held state enumeration could not reach.
In 0.3 the store was flat *and* enumerable, which is why this only shows
up now.

## The layer this is fixed at

The backend, where paths are resolved — not per command. `Memory` now
resolves a path one component at a time, following symlinks at every
step:

- a non-final component that resolves to a **regular file** is `:enotdir`
  (the backend names the offending component; vfs rewrites the error path
  to the requested path, which is what bash prints);
- a non-final component that is **missing** is still created — writes
  create their parent directories, as before;
- a non-final component that is a **symlink to a directory** now resolves
  to the target directory, instead of storing a literal `/link/sub/a.md`
  key that no listing could reach (the same bug class);
- a symlink cycle anywhere in the path is `:eloop`.

Two entry points: `resolve_path/2` follows a symlink at the final
component too (`O_TRUNC`/`O_APPEND`/`chmod` semantics), `resolve_for_create/2`
takes the final component literally (`mkdir`/`symlink`/`link`, where an
existing final component is `EEXIST`, not something to follow; and
`lstat`/`readlink`/`rm`, where the link itself is the subject). This
replaces the old `resolve_final_path/3`, which only ever looked at the
last component.

Both are one walk over the path's components, carrying the resolved prefix
as an accumulator. Symlink hops are budgeted (`SYMLOOP_MAX` 40) rather than
remembered: a set of visited links rejects legitimate paths that cross the
same link twice — with `/real/self -> /real`, `/real/self/self/x` is two
hops to `/real/x`, not a cycle.

`write_file/4`, `append_file/3`, `mkdir/3`, `symlink/3`, `link/3`, and
`chmod/3` all go through it, so every caller — the redirection layer and
every command — gets the check for free. Reads resolve through the same walk, and
have to: if only writes followed intermediate symlinks, a write through a
symlinked directory would land on the target and be invisible at the path
the caller used — the same unreachable-state bug mirrored, and quieter,
because the write reports success. So `stat`, `readdir`, `stream_read`,
`exists?`, `rm`, `lstat`, `readlink`, and `link`'s source path all route
through it, and `echo hi > /link/a.md` is readable back at `/link/a.md`.

One deviation remains: `normalize/1` collapses `..` lexically before
resolution, so `echo hi > /m/j/../k.md` writes `/m/k.md` and exits 0 where
bash reports `Not a directory`. It does not admit unreachable state, so it
stayed out of scope — UPGRADING item 7 states it.

## Behaviors that changed

With `/m/j` a regular file, all of these were exit 0 (or crashed) and are
now exit 1 with bash's wording — verified against real bash:

| command | now |
|---|---|
| `echo hi > /m/j/a.md` | `bash: /m/j/a.md: Not a directory` |
| `echo hi > /m/j/2026/a.md` | `bash: /m/j/2026/a.md: Not a directory` |
| `echo hi >> …`, `&> …`, `2> …` | same, per stream |
| `mkdir /m/j/2026` | `mkdir: cannot create directory '/m/j/2026': Not a directory` |
| `mkdir -p /m/j/2026` | `mkdir: cannot create directory '/m/j': Not a directory` (GNU names the offending ancestor for `-p`) |
| `touch /m/j/a.md` | `touch: cannot touch '/m/j/a.md': Not a directory` |
| `cp /src.txt /m/j/a.md` | `cp: cannot stat '/m/j/a.md': Not a directory` (GNU stats the destination first; previously a `MatchError` crash out of `JustBash.exec/2`) |
| `mv /src.txt /m/j/a.md` | `mv: cannot move '/src.txt' to '/m/j/a.md': Not a directory`, source kept |
| `echo hi \| tee /m/j/a.md` | `tee: /m/j/a.md: Not a directory`, stdin still passed through |
| `ln -s /src.txt /m/j/link` | `ln: failed to create symbolic link '/m/j/link': Not a directory` |

Adjacent fixes the tests turned up:

- **`cp`** matched `{:ok, new_fs} = FS.write_file(...)`, so *any*
  unwritable destination crashed the exec. It reports now.
- **`touch`** and **`tee`** interpolated a fixed reason regardless of the
  actual error (`touch: cannot touch 'X'` with no reason at all; `tee: X:
  Is a directory` for every write failure). Both render `FS.strerror/1`,
  matching precedent dffd5dd.
- **`mkdir -p`** swallowed `:eexist` unconditionally, so `mkdir -p
  /m/j` with `/m/j` a regular file exited 0 while no such directory
  existed. It now requires the existing entry to be a directory:
  `mkdir: cannot create directory '/m/j': File exists`, exit 1 (bash
  agrees). A dangling symlink reports the same, via an `FS.lstat` fallback
  when `FS.stat` cannot resolve the name.
- **`mv`**'s error `case` had no fallback clause and would have raised
  `CaseClauseError` on any kind it did not enumerate.

## `JustBash.new/1` with an inconsistent file map

Decided: **raise**. `%{"/m/j" => "x", "/m/j/a.md" => "y"}` describes a
filesystem that cannot exist, and silently accepting it was doubly bad —
which entry survived depended on map iteration order (one order stored an
unreachable child, the other lost the child to an `:eisdir` write). It is
a caller mistake at construction time, so `JustBash.new/1`,
`JustBash.FS.new/1`, and `JustBash.FS.Memory.new/1` now raise
`ArgumentError` naming both paths:

```
invalid initial files: "/m/j" is a regular file, so "/m/j/a.md" cannot
exist under it (POSIX path resolution fails with ENOTDIR). Drop one of
the two entries.
```

The check compares path *components*, so `%{"/m/jj" => …, "/m/j" => …}`
is still fine. `JustBash.new/1` also turns a write failure against the
default directories (`files: %{"/tmp" => "x"}`) into the same
`ArgumentError` instead of a `MatchError` from deep inside
`init_filesystem/1`.

## Tests

66 tests in `test/just_bash/not_a_directory_test.exs`, written
test-first (31 failed on the first run of the original 37; the 29 added in
the review round took it to 30 of 66 failing before the fixes): FS-layer `:enotdir` for every path-creating operation, the
walk/read reachability invariant, the symlinked-directory resolution that
now works, bash-level stderr and exit codes for each command above, and
the inconsistent-map raises. `UPGRADING.md` gets the bash-level delta as
item 7 of the unreleased 0.3 → 0.4 list.

All five checks pass: `mix compile --warnings-as-errors`, `mix dialyzer`
(0 new errors; the one pre-existing unnecessary skip is unchanged from
main — and `memory.ex` now needs no `@dialyzer` attribute at all),
`mix credo --strict` (only the intentional test-fixture finding),
`mix format --check-formatted`, `mix test` — 3,806 tests, 0 failures.

## Review round (c868a1e)

The first push enforced ENOTDIR on the write side but left reads resolving
literal paths, which regressed writes through a symlinked directory — they
succeeded and were then unreadable at the path used. Fixed by sharing the
walk, which also closed three other findings. Alongside it:

- **Perf.** `mkdir_with_parent/3` was recursing through the `VFS.Mountable`
  protocol, re-resolving an already-validated chain once per level; it calls
  `do_mkdir_resolved/3` directly now. With the single-pass walk, `mkdir -p`
  at depth 400 went 698 ms → 4.2 ms, *faster* than before this PR (6.6 ms).
  Writes at depth 12 went 187 ms → 48 ms against a 34 ms baseline; reads pay
  ~3%.
- **GNU message parity.** `mkdir -p` names the offending ancestor in the
  operand's own form; `cp` uses `cannot stat` for ENOTDIR; `tee` one level
  deeper says `Not a directory`.
- **A failed redirect no longer leaks the command's stdout.** The bytes were
  bound for the file, so `echo hi > /m/j/a.md` yields empty stdout. bash
  produces nothing because it opens the target before running the command;
  clearing the stream is how close this gets without pre-flighting.
- **`Memory.new/1`** raises `ArgumentError` on any unwritable seed
  (`%{"/" => "x"}`), so all three entry points agree and the docstrings are
  true.
- **`link/3`** resolves its source path's ancestors too.

`mv`'s pre-existing `cannot move 'X' to 'Y'` wording is left alone — #57
builds on it, and changing it here would make that merge produce the wrong
text.

### DIFF
--- a/lib/just_bash/commands/cp.ex
+++ b/lib/just_bash/commands/cp.ex
@@ -17,8 +17,7 @@ defmodule JustBash.Commands.Cp do
 
         case FS.read_file(bash.fs, src_resolved) do
           {:ok, content, fs} ->
-            {:ok, new_fs} = FS.write_file(fs, dest_resolved, content)
-            {Command.ok(), %{bash | fs: new_fs}}
+            write_dest(%{bash | fs: fs}, dest_resolved, dest, content)
 
           {:error, _} ->
             {Command.error("cp: cannot stat '#{src}': No such file or directory\n"), bash}
@@ -26,6 +25,20 @@ defmodule JustBash.Commands.Cp do
 
       _ ->
         {Command.error("cp: missing file operand\n"), bash}
+    end
+  end
+
+  # The destination may be unwritable for reasons the source read cannot
+  # know about — a path component that is a regular file (:enotdir), a
+  # destination that is a directory (:eisdir). Report them instead of
+  # crashing on the write.
+  defp write_dest(bash, dest_resolved, dest, content) do
+    case FS.write_file(bash.fs, dest_resolved, content) do
+      {:ok, new_fs} ->
+        {Command.ok(), %{bash | fs: new_fs}}
+
+      {:error, %VFS.Error{} = error} ->
+        {Command.error("cp: cannot create regular file '#{dest}': #{FS.strerror(error)}\n"), bash}
     end
   end
 end
--- a/lib/just_bash/commands/cp.ex
+++ b/lib/just_bash/commands/cp.ex
@@ -37,6 +37,13 @@ defmodule JustBash.Commands.Cp do
       {:ok, new_fs} ->
         {Command.ok(), %{bash | fs: new_fs}}
 
+      # GNU stats the destination before opening it, so a path component that
+      # is a regular file surfaces as `cannot stat`. `cannot create regular
+      # file` stays the wording for a destination whose parent is merely
+      # missing.
+      {:error, %VFS.Error{kind: :enotdir} = error} ->
+        {Command.error("cp: cannot stat '#{dest}': #{FS.strerror(error)}\n"), bash}
+
       {:error, %VFS.Error{} = error} ->
         {Command.error("cp: cannot create regular file '#{dest}': #{FS.strerror(error)}\n"), bash}
     end
--- a/lib/just_bash/commands/mkdir.ex
+++ b/lib/just_bash/commands/mkdir.ex
@@ -23,9 +23,16 @@ defmodule JustBash.Commands.Mkdir do
           {:error, %VFS.Error{kind: :eexist}} when flags.p ->
             handle_existing_path(fs_acc, resolved, path, err_acc, code_acc)
 
+          # With -p, the directory GNU reports is the shallowest one it could
+          # not create — the offending ancestor — not the whole operand:
+          #   mkdir j/2026           -> cannot create directory 'j/2026'
+          #   mkdir -p j/2026/deeper -> cannot create directory 'j'
+          {:error, %VFS.Error{kind: :enotdir} = error} when flags.p ->
+            named = offending_ancestor(fs_acc, bash.cwd, path)
+            {err_acc <> failure(named, error), 1, fs_acc}
+
           {:error, %VFS.Error{} = error} ->
-            {err_acc <> "mkdir: cannot create directory '#{path}': #{FS.strerror(error)}\n", 1,
-             fs_acc}
+            {err_acc <> failure(path, error), 1, fs_acc}
         end
       end)
 
@@ -43,9 +50,51 @@ defmodule JustBash.Commands.Mkdir do
       {:ok, _stat, new_fs} ->
         {err_acc <> "mkdir: cannot create directory '#{path}': File exists\n", 1, new_fs}
 
+      # `stat` fails on a dangling symlink, but the *name* is taken, so
+      # `mkdir` cannot have it — GNU says "File exists". Only a name that is
+      # absent in `lstat` too gets the stat error.
       {:error, %VFS.Error{} = error} ->
-        {err_acc <> "mkdir: cannot create directory '#{path}': #{FS.strerror(error)}\n", 1, fs}
+        case FS.lstat(fs, resolved) do
+          {:ok, _stat, new_fs} ->
+            {err_acc <> "mkdir: cannot create directory '#{path}': File exists\n", 1, new_fs}
+
+          {:error, %VFS.Error{}} ->
+            {err_acc <> failure(path, error), 1, fs}
+        end
     end
+  end
+
+  # Walk the operand's own components so the name is reported the way the
+  # caller wrote it ("j", not "/m/j"), and stop at the first one that exists
+  # but is not a directory.
+  defp offending_ancestor(fs, cwd, path) do
+    trimmed = String.trim_trailing(path, "/")
+
+    trimmed
+    |> operand_prefixes()
+    |> Enum.find(&non_directory?(fs, FS.resolve_path(cwd, &1)))
+    |> Kernel.||(path)
+  end
+
+  defp operand_prefixes(path) do
+    case String.split(path, "/") do
+      ["" | rest] -> rest |> prefixes() |> Enum.map(&("/" <> &1))
+      components -> prefixes(components)
+    end
+  end
+
+  defp prefixes(components) do
+    components
+    |> Enum.reject(&(&1 == ""))
+    |> Enum.scan(&Path.join(&2, &1))
+  end
+
+  defp non_directory?(fs, resolved) do
+    match?({:ok, %VFS.Stat{type: type}, _fs} when type != :directory, FS.stat(fs, resolved))
+  end
+
+  defp failure(path, error) do
+    "mkdir: cannot create directory '#{path}': #{FS.strerror(error)}\n"
   end
 
   defp parse_flags(args), do: parse_flags(args, %{p: false}, [])
--- a/lib/just_bash/commands/mkdir.ex
+++ b/lib/just_bash/commands/mkdir.ex
@@ -21,10 +21,7 @@ defmodule JustBash.Commands.Mkdir do
             {err_acc, code_acc, new_fs}
 
           {:error, %VFS.Error{kind: :eexist}} when flags.p ->
-            {err_acc, code_acc, fs_acc}
-
-          {:error, %VFS.Error{kind: :eexist}} ->
-            {err_acc <> "mkdir: cannot create directory '#{path}': File exists\n", 1, fs_acc}
+            handle_existing_path(fs_acc, resolved, path, err_acc, code_acc)
 
           {:error, %VFS.Error{} = error} ->
             {err_acc <> "mkdir: cannot create directory '#{path}': #{FS.strerror(error)}\n", 1,
@@ -33,6 +30,22 @@ defmodule JustBash.Commands.Mkdir do
       end)
 
     {Command.result("", stderr, exit_code), %{bash | fs: new_fs}}
+  end
+
+  # `mkdir -p` is only satisfied by an existing *directory*. A regular file
+  # at the path is still an error (bash: "mkdir: PATH: File exists"), and
+  # swallowing it would report success for a directory that does not exist.
+  defp handle_existing_path(fs, resolved, path, err_acc, code_acc) do
+    case FS.stat(fs, resolved) do
+      {:ok, %VFS.Stat{type: :directory}, new_fs} ->
+        {err_acc, code_acc, new_fs}
+
+      {:ok, _stat, new_fs} ->
+        {err_acc <> "mkdir: cannot create directory '#{path}': File exists\n", 1, new_fs}
+
+      {:error, %VFS.Error{} = error} ->
+        {err_acc <> "mkdir: cannot create directory '#{path}': #{FS.strerror(error)}\n", 1, fs}
+    end
   end
 
   defp parse_flags(args), do: parse_flags(args, %{p: false}, [])
--- a/lib/just_bash/commands/tee.ex
+++ b/lib/just_bash/commands/tee.ex
@@ -64,6 +64,9 @@ defmodule JustBash.Commands.Tee do
     end)
   end
 
+  # tee does not create missing parent directories, so an absent parent is
+  # reported instead of being conjured up; a parent that exists but is not
+  # a directory is ENOTDIR, as the kernel would report it.
   defp write_single_file(fs, resolved, file, content, append, acc_stderr, acc_code) do
     parent = Path.dirname(resolved)
 
@@ -72,19 +75,20 @@ defmodule JustBash.Commands.Tee do
         do_write_file(new_fs, resolved, file, content, append, acc_stderr, acc_code)
 
       {:ok, _, new_fs} ->
-        {new_fs, acc_stderr <> "tee: #{file}: No such file or directory\n", 1}
+        {new_fs, acc_stderr <> "tee: #{file}: Not a directory\n", 1}
 
-      {:error, _} ->
-        {fs, acc_stderr <> "tee: #{file}: No such file or directory\n", 1}
+      {:error, %VFS.Error{} = error} ->
+        {fs, acc_stderr <> "tee: #{file}: #{FS.strerror(error)}\n", 1}
     end
   end
 
   defp do_write_file(fs, resolved, file, content, append, acc_stderr, acc_code) do
-    result = write_or_append(fs, resolved, content, append)
+    case write_or_append(fs, resolved, content, append) do
+      {:ok, new_fs} ->
+        {new_fs, acc_stderr, acc_code}
 
-    case result do
-      {:ok, new_fs} -> {new_fs, acc_stderr, acc_code}
-      {:error, _} -> {fs, acc_stderr <> "tee: #{file}: Is a directory\n", 1}
+      {:error, %VFS.Error{} = error} ->
+        {fs, acc_stderr <> "tee: #{file}: #{FS.strerror(error)}\n", 1}
     end
   end
 
--- a/lib/just_bash/fs/fs.ex
+++ b/lib/just_bash/fs/fs.ex
@@ -40,7 +40,9 @@ defmodule JustBash.FS do
   backend (seeded with `initial_files`) mounted at `/`.
 
   Raises `ArgumentError` when `initial_files` is not realizable as a
-  filesystem — see `validate_initial_files!/1`.
+  filesystem: either one entry's path runs through another (see
+  `validate_initial_files!/1`), or an entry collides with a directory the
+  backend already holds, as `%{"/" => "x"}` does.
   """
   @spec new(map()) :: t()
   def new(initial_files \\ %{}) do
--- a/lib/just_bash/fs/fs.ex
+++ b/lib/just_bash/fs/fs.ex
@@ -38,11 +38,25 @@ defmodule JustBash.FS do
   @doc """
   Create a new filesystem: a `%VFS{}` with a `JustBash.FS.Memory`
   backend (seeded with `initial_files`) mounted at `/`.
+
+  Raises `ArgumentError` when `initial_files` is not realizable as a
+  filesystem — see `validate_initial_files!/1`.
   """
   @spec new(map()) :: t()
   def new(initial_files \\ %{}) do
     VFS.new() |> VFS.mount("/", Memory.new(initial_files))
   end
+
+  @doc """
+  Validate a `path => content` map before seeding a filesystem with it.
+
+  Raises `ArgumentError` when one entry's path runs *through* another
+  entry — `%{"/m/j" => "x", "/m/j/a.md" => "y"}` asks for a file inside a
+  regular file, which no filesystem can hold. See
+  `JustBash.FS.Memory.validate_initial_files!/1`.
+  """
+  @spec validate_initial_files!(map()) :: :ok
+  defdelegate validate_initial_files!(initial_files), to: Memory
 
   # ── path helpers (pure) ──────────────────────────────────────────────────
   #
--- a/lib/just_bash/fs/memory.ex
+++ b/lib/just_bash/fs/memory.ex
@@ -77,8 +77,10 @@ defmodule JustBash.FS.Memory do
   - Extended form: `%{"/path/to/file" => %{content: "content", mode: 0o755, mtime: ~U[...]}}`
 
   Parent directories are created automatically. Raises `ArgumentError`
-  when the map is not realizable as a filesystem — see
-  `validate_initial_files!/1`.
+  when the map is not realizable as a filesystem: either one entry's path
+  runs through another (see `validate_initial_files!/1`), or an entry
+  collides with a directory this backend already holds, as `%{"/" => "x"}`
+  does.
 
   ## Examples
 
@@ -97,19 +99,31 @@ defmodule JustBash.FS.Memory do
     Enum.reduce(initial_files, fs, fn {path, value}, acc ->
       case value do
         %{content: content} = init ->
-          {:ok, new_fs} =
-            write_file(acc, path, content,
-              mode: Map.get(init, :mode, 0o644),
-              mtime: Map.get(init, :mtime, DateTime.utc_now())
-            )
-
-          new_fs
+          seed!(acc, path, content,
+            mode: Map.get(init, :mode, 0o644),
+            mtime: Map.get(init, :mtime, DateTime.utc_now())
+          )
 
         content when is_binary(content) ->
-          {:ok, new_fs} = write_file(acc, path, content)
-          new_fs
+          seed!(acc, path, content)
       end
     end)
+  end
+
+  # `validate_initial_files!/1` catches the map that is unrealizable on its
+  # own terms; a seed can still collide with what the backend already holds
+  # (`%{"/" => "x"}`). Both are caller mistakes at construction time, so
+  # both raise `ArgumentError` rather than a `MatchError` from in here.
+  defp seed!(fs, path, content, opts \\ []) do
+    case write_file(fs, path, content, opts) do
+      {:ok, new_fs} ->
+        new_fs
+
+      {:error, %Error{} = error} ->
+        raise ArgumentError,
+              "invalid initial files: cannot create #{inspect(path)}: " <>
+                JustBash.FS.strerror(error)
+    end
   end
 
   @doc """
@@ -155,7 +169,7 @@ defmodule JustBash.FS.Memory do
   def write_file(%__MODULE__{} = fs, path, content, opts \\ []) do
     normalized = normalize(path)
 
-    with {:ok, target_path} <- resolve_for_write(fs, normalized) do
+    with {:ok, target_path} <- resolve_path(fs, normalized) do
       case Map.get(fs.data, target_path) do
         %{type: :directory} ->
           {:error, Error.new(:eisdir, path: normalized)}
@@ -174,22 +188,24 @@ defmodule JustBash.FS.Memory do
   @doc """
   Get stat information for a path without following symlinks.
 
-  A symlink reports `type: :symlink` with the target's byte size.
+  A symlink reports `type: :symlink` with the target's byte size. Ancestor
+  components still resolve — only the final one is left alone.
   """
   @spec lstat(t(), String.t()) :: {:ok, Stat.t(), t()} | {:error, Error.t()}
-  def lstat(%__MODULE__{data: data} = fs, path) do
-    normalized = normalize(path)
+  def lstat(%__MODULE__{} = fs, path) do
+    with {:ok, normalized} <- resolve_for_create(fs, normalize(path)) do
+      case Map.get(fs.data, normalized) do
+        nil ->
+          {:error, Error.new(:enoent, path: normalized)}
 
-    case Map.get(data, normalized) do
-      nil ->
-        {:error, Error.new(:enoent, path: normalized)}
+        %{type: :symlink, target: target} = entry ->
+          {:ok,
+           %Stat{type: :symlink, size: byte_size(target), mtime: entry.mtime, mode: entry.mode},
+           fs}
 
-      %{type: :symlink, target: target} = entry ->
-        {:ok,
-         %Stat{type: :symlink, size: byte_size(target), mtime: entry.mtime, mode: entry.mode}, fs}
-
-      entry ->
-        {:ok, entry_stat(entry), fs}
+        entry ->
+          {:ok, entry_stat(entry), fs}
+      end
     end
   end
 
@@ -215,15 +231,17 @@ defmodule JustBash.FS.Memory do
 
   @doc """
   Read the target of a symbolic link.
+
+  Ancestor components resolve; the final one is the link being read.
   """
   @spec readlink(t(), String.t()) :: {:ok, String.t(), t()} | {:error, Error.t()}
-  def readlink(%__MODULE__{data: data} = fs, path) do
-    normalized = normalize(path)
-
-    case Map.get(data, normalized) do
-      nil -> {:error, Error.new(:enoent, path: normalized)}
-      %{type: :symlink, target: target} -> {:ok, target, fs}
-      _ -> {:error, Error.new(:einval, path: normalized)}
+  def readlink(%__MODULE__{} = fs, path) do
+    with {:ok, normalized} <- resolve_for_create(fs, normalize(path)) do
+      case Map.get(fs.data, normalized) do
+        nil -> {:error, Error.new(:enoent, path: normalized)}
+        %{type: :symlink, target: target} -> {:ok, target, fs}
+        _ -> {:error, Error.new(:einval, path: normalized)}
+      end
     end
   end
 
@@ -237,8 +255,9 @@ defmodule JustBash.FS.Memory do
   """
   @spec link(t(), String.t(), String.t()) :: {:ok, t()} | {:error, Error.t()}
   def link(%__MODULE__{} = fs, existing_path, new_path) do
-    with {:ok, new_norm} <- resolve_for_create(fs, normalize(new_path)) do
-      do_link(fs, normalize(existing_path), new_norm)
+    with {:ok, new_norm} <- resolve_for_create(fs, normalize(new_path)),
+         {:ok, existing_norm} <- resolve_for_create(fs, normalize(existing_path)) do
+      do_link(fs, existing_norm, new_norm)
     end
   end
 
@@ -271,7 +290,7 @@ defmodule JustBash.FS.Memory do
   def chmod(%__MODULE__{} = fs, path, mode) do
     normalized = normalize(path)
 
-    with {:ok, target_path} <- resolve_for_write(fs, normalized) do
+    with {:ok, target_path} <- resolve_path(fs, normalized) do
       case Map.get(fs.data, target_path) do
         nil ->
           {:error, Error.new(:enoent, path: normalized)}
@@ -296,7 +315,7 @@ defmodule JustBash.FS.Memory do
   def append_file(%__MODULE__{} = fs, path, content) do
     normalized = normalize(path)
 
-    with {:ok, target_path} <- resolve_for_write(fs, normalized) do
+    with {:ok, target_path} <- resolve_path(fs, normalized) do
       case Map.get(fs.data, target_path) do
         %{type: :directory} ->
           {:error, Error.new(:eisdir, path: normalized)}
@@ -318,10 +337,20 @@ defmodule JustBash.FS.Memory do
   def __entry__(%__MODULE__{data: data}, path), do: Map.get(data, path)
 
   @doc false
-  @spec __resolve_entry__(t(), String.t()) ::
-          {:ok, fs_entry()} | {:error, :enoent | :eloop}
+  @spec __resolve_entry__(t(), String.t()) :: {:ok, fs_entry()} | {:error, Error.t()}
   def __resolve_entry__(%__MODULE__{} = fs, path) do
-    do_resolve_entry(fs, normalize(path), MapSet.new())
+    with {:ok, resolved} <- resolve_path(fs, normalize(path)) do
+      case Map.get(fs.data, resolved) do
+        nil -> {:error, Error.new(:enoent, path: resolved)}
+        entry -> {:ok, entry}
+      end
+    end
+  end
+
+  @doc false
+  @spec __resolve_for_read__(t(), String.t()) :: {:ok, String.t()} | {:error, Error.t()}
+  def __resolve_for_read__(%__MODULE__{} = fs, path) do
+    resolve_path(fs, normalize(path))
   end
 
   @doc false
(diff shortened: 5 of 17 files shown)


lib/just_bash/fs/memory.ex
```diff
@@ -361,25 +391,100 @@ defmodule JustBash.FS.Memory do
...
+      {:ok, join_child(parent, VPath.basename(path))}
+    end
+  end
+
+  defp resolve_directory(%__MODULE__{} = fs, path, seen) do
+    with {:ok, resolved} <- resolve_full(fs, path, seen) do
+      case Map.get(fs.data, resolved) do
+        nil -> {:ok, resolved}
+        %{type: :directory} -> {:ok, resolved}
+        _ -> {:error, Error.new(:enotdir, path: resolved)}
```
davydog187: **Blocking — the read side does not do this, so a write through a symlinked directory becomes unreachable at the path used.**

`resolve_directory/3` correctly resolves `/link` to `/real`, so `write_file("/link/a.md")` stores the entry at `/real/a.md`. But the `VFS.Mountable` read side (`stat/2`:507, `readdir/2`:516, `stream_read/3`:544, `exists?/2`:503, `rm/3`:576) and `Memory.lstat/2`:180 / `readlink/2`:220 all do `Map.get(data, normalize(path))` on the **literal** path, and `__resolve_entry__/2`:323 only follows a symlink at the final component. So:

```
mkdir -p /real && ln -s /real /link
echo hi > /link/a.md   # exit 0
cat /link/a.md         # exit 1, ENOENT   (main: "hi", bash: "hi")
rm /link/f             # exit 1           (main: 0,   bash: 0)
mkdir -p /link/sub; ls /link/sub  # exit 1 (main: 0,  bash: 0)
```

All verified regressions against `56c74b8`. Suggest a `resolve_for_read` built on the same walk, used by `stat`/`readdir`/`stream_read`/`exists?`, plus ancestor-only resolution (i.e. `resolve_for_create`) for `lstat`/`readlink`/`rm`, which must not follow the final link.

lib/just_bash/fs/memory.ex
```diff
@@ -361,25 +391,100 @@ defmodule JustBash.FS.Memory do
...
+  # file — readable by its path, yet unreachable from any directory
+  # listing or `VFS.walk/3` (issue #53).
+
+  # Resolve every component *including* a symlink at the final one:
+  # O_TRUNC/O_APPEND/chmod semantics, where the link is written through. A
+  # missing entry resolves to the path it would occupy — that is where a
+  # write through a dangling link creates the file.
+  # The `seen` MapSet threads through the mutually recursive walkers, and
+  # dialyzer cannot see through the opaque struct across the recursion.
+  @dialyzer {:nowarn_function,
```
davydog187: Nit, verified: three of these four are unnecessary. Narrowing the list to `[follow_link: 3]` gives `mix dialyzer` -> `Total errors: 17, Skipped: 17, Unnecessary Skips: 1`, identical to this branch as submitted. Removing the attribute entirely produces exactly two warnings, both inside `follow_link/3`:

```
memory.ex:449:27 call_without_opaque  Type mismatch in call without opaque term in member?.
memory.ex:453:45 call_with_opaque     Type mismatch in call with opaque term in resolve_full.
```

So `resolve_full: 3`, `resolve_ancestors: 3` and `resolve_directory: 3` can come off, and the trailing comment ("dialyzer cannot see through the opaque struct across the recursion") becomes precisely true of the one function it is describing.

lib/just_bash/fs/memory.ex
```diff
@@ -497,6 +593,22 @@ defimpl VFS.Mountable, for: JustBash.FS.Memory do
...
+        {:error, Error.new(:eexist, path: normalized)}
+
+      %{type: :directory} when not parents? ->
+        {:error, Error.new(:eexist, path: normalized)}
+
+      %{type: :directory} ->
+        {:ok, fs}
+
+      nil ->
+        mkdir_with_parent(fs, normalized, parents?)
```
davydog187, line 738: **Perf: `mkdir -p` is now up to ~97x slower on deep paths, and this is the cause.**

```
mkdir -p depth 100:   main 0.5 ms  ->  PR  12 ms   (22x)
mkdir -p depth 200:   main 1.8 ms  ->  PR  86 ms   (47x)
mkdir -p depth 400:   main 7.2 ms  ->  PR 698 ms   (97x)
```

`mkdir_with_parent/3` recurses through the **protocol** `mkdir/3` (line 621), which calls `Memory.__resolve_for_create__/2` again at every level. The chain has already been fully validated by the outer `mkdir/3`, so this re-resolves an already-resolved path n times, each resolution doing O(n) `dirname`/`basename`/`join_child` string allocations — O(n^3) bytes overall.

Calling `mkdir_with_parent/3` (or `do_mkdir_resolved/3`) directly on the already-resolved parent, instead of going back through the protocol, should restore roughly the old cost.

Separately, for the per-write cost (10k writes at depth 12: 35 ms -> 187 ms), `resolve_full/3` recursing on `dirname` takes the path apart and rebuilds it at every level. Splitting once into components and walking down with an accumulator would make resolution a single pass and keep it O(path length) rather than O(depth x path length).

lib/just_bash/commands/mkdir.ex
```diff
@@ -35,6 +32,22 @@ defmodule JustBash.Commands.Mkdir do
     {Command.result("", stderr, exit_code), %{bash | fs: new_fs}}
   end
 
+  # `mkdir -p` is only satisfied by an existing *directory*. A regular file
+  # at the path is still an error (bash: "mkdir: PATH: File exists"), and
+  # swallowing it would report success for a directory that does not exist.
+  defp handle_existing_path(fs, resolved, path, err_acc, code_acc) do
+    case FS.stat(fs, resolved) do
```
davydog187, line 46: Two notes on this helper. The main behaviour is right — `mkdir -p` on a symlink-to-directory still exits 0 and on a regular file now correctly says `File exists`, matching GNU. But:

**1. `mkdir -p` names the wrong path in the ENOTDIR message.** GNU coreutils 9.11 names the offending *ancestor* for `-p`, and the full operand without it:

```
$ gmkdir j/2026            -> gmkdir: cannot create directory 'j/2026': Not a directory
$ gmkdir -p j/2026         -> gmkdir: cannot create directory 'j': Not a directory
$ gmkdir -p j/2026/deeper  -> gmkdir: cannot create directory 'j': Not a directory
```

This PR emits `cannot create directory '/m/j/2026': Not a directory` for both, and the test at `not_a_directory_test.exs:824` pins it. #53's own table also expects `mkdir: /m/j: Not a directory`. The backend does know the offending component (`Error.new(:enotdir, path: resolved)`), but `VFS.Error.put_path/2` overwrites it with the requested path on the way out, so surfacing it needs the component carried separately.

**2. Dangling symlink reports the wrong errno.** `FS.stat` fails on a dangling link, so this falls to the `{:error, ...}` clause:

```
ln -s /nowhere /dang; mkdir -p /dang
  PR:   mkdir: cannot create directory '/dang': No such file or directory
  GNU:  mkdir: cannot create directory 'dang': File exists
  main: exit 0 (silently wrong)
```

Still an improvement over `main`, just not the right message. Falling back to `FS.lstat` on stat failure — the name exists, so `File exists` — would cover it.

lib/just_bash/commands/cp.ex
```diff
@@ -28,4 +27,18 @@ defmodule JustBash.Commands.Cp do
...
+  # know about — a path component that is a regular file (:enotdir), a
+  # destination that is a directory (:eisdir). Report them instead of
+  # crashing on the write.
+  defp write_dest(bash, dest_resolved, dest, content) do
+    case FS.write_file(bash.fs, dest_resolved, content) do
+      {:ok, new_fs} ->
+        {Command.ok(), %{bash | fs: new_fs}}
+
+      {:error, %VFS.Error{} = error} ->
+        {Command.error("cp: cannot create regular file '#{dest}': #{FS.strerror(error)}\n"), bash}
```
davydog187, line 48: GNU stats the destination first, so ENOTDIR surfaces as `cannot stat`, not `cannot create regular file`:

```
$ gcp src.txt j/a.md  -> gcp: cannot stat 'j/a.md': Not a directory
$ gmv src.txt j/a.md  -> gmv: cannot stat 'j/a.md': Not a directory
```

`cannot create regular file` is right for `:enoent` (missing parent), wrong for `:enotdir`. Same applies to `mv.ex:47`'s `cannot move '#{src}' to '#{dest}'` — though that wording is pre-existing there, whereas this one is new and `not_a_directory_test.exs:865` pins it.

Note this cannot be fixed here alone: `cp` cannot distinguish the two until `FS.stat` on a path under a regular file returns `:enotdir` instead of `:enoent` (see the read-side comment on `memory.ex`).

The fix in this hunk itself is unambiguously right, by the way — the old `{:ok, new_fs} = FS.write_file(...)` let a `MatchError` escape `JustBash.exec/2` for *any* unwritable destination.

lib/just_bash/commands/tee.ex
```diff
@@ -72,19 +75,20 @@ defmodule JustBash.Commands.Tee do
         do_write_file(new_fs, resolved, file, content, append, acc_stderr, acc_code)
 
       {:ok, _, new_fs} ->
-        {new_fs, acc_stderr <> "tee: #{file}: No such file or directory\n", 1}
+        {new_fs, acc_stderr <> "tee: #{file}: Not a directory\n", 1}
```
davydog187, line 78: This is correct for a parent that exists as a regular file, and matches GNU exactly. But it only covers the shallow case, because it depends on `FS.stat` finding the parent — and one level deeper the read side returns `:enoent`, so the `{:error, ...}` clause below renders the wrong errno:

```
echo hi | tee /m/j/a.md       -> tee: /m/j/a.md: Not a directory            (matches GNU)
echo hi | tee /m/j/2026/a.md  -> tee: /m/j/2026/a.md: No such file or directory
                                 GNU: tee: j/2026/a.md: Not a directory
```

The new tests only cover the shallow form, so this is invisible in the suite. It resolves for free once the read side does component-wise resolution.

test/just_bash/not_a_directory_test.exs
```diff
@@ -0,0 +1,316 @@
...
+      assert {:ok, "x", _fs} = FS.read_file(fs, "/src.txt")
+    end
+  end
+
+  describe "shell redirections through a regular file" do
+    test "> reports Not a directory when the parent is a regular file" do
+      {result, bash} = JustBash.exec(bash_with_file_parent(), "echo hi > /m/j/a.md")
+
+      assert result.exit_code == 1
+      assert result.stderr == "bash: /m/j/a.md: Not a directory\n"
```
davydog187, line 159: These four redirect tests assert `exit_code` and `stderr` but not `stdout`, which hides a real observable difference: the failed redirect still leaks the command's output to the parent's stdout.

```elixir
JustBash.exec(bash_with_file_parent(), "echo hi > /m/j/a.md")
#=> exit_code: 1, stderr: "bash: /m/j/a.md: Not a directory\n", stdout: "hi\n"
#   real bash: no stdout at all — the redirect fails before the command runs
```

Same on `main` for `echo hi > /somedir` (`:eisdir`), so this PR did not introduce it. But the PR converts a large class of previously-successful writes into failures, so the leak goes from an exotic path to a common one, and for a sandboxed caller it means the payload lands in the stdout they read back.

Adding `assert result.stdout == ""` to these four documents the intent and turns a silent divergence into a tracked one.

lib/just_bash/fs/fs.ex
```diff
@@ -38,12 +38,26 @@ defmodule JustBash.FS do
   @doc """
   Create a new filesystem: a `%VFS{}` with a `JustBash.FS.Memory`
   backend (seeded with `initial_files`) mounted at `/`.
+
+  Raises `ArgumentError` when `initial_files` is not realizable as a
```
davydog187, line 42: This docstring (and the matching one on `Memory.new/1`) overstates what actually raises. Only the nested-path case is validated; an impossible map that is not a path-through-path still raises `MatchError` from `Memory.new/1`'s `{:ok, new_fs} = write_file(...)`:

```elixir
iex> JustBash.FS.new(%{"/" => "x"})
** (MatchError) no match of right hand side value:
    {:error, %VFS.Error{kind: :eisdir, path: "/", ...}}
    (just_bash) lib/just_bash/fs/memory.ex:109

iex> JustBash.new(files: %{"/" => "x"})
** (ArgumentError) invalid initial files: cannot create "/": Is a directory
```

`JustBash.new/1` gets it right because `init_filesystem/1` wraps the write. Moving that same wrapping into `Memory.new/1`'s reduce would make all three entry points agree and make the docstring true — otherwise it needs softening to "raises `ArgumentError` when one path runs through another".

davydog187 (commented):
## Review

The diagnosis in the description is right and the layer chosen is right: enforcing POSIX component resolution in the backend, so the redirection layer and every command inherit the check, is much better than patching each command. The `resolve_for_write` / `resolve_for_create` split (follow the final symlink for `O_TRUNC`/`O_APPEND`/`chmod`, take it literally for `mkdir`/`symlink`/`link`) is the correct distinction and is applied correctly. The adjacent fixes are real: `cp`'s hard match and `mv`'s missing `case` fallback were both live crash paths out of `JustBash.exec/2`.

I verified all four tooling claims locally on `d00f5c4`: `mix test` 3776/0, `mix format --check-formatted` clean, `mix credo --strict` clean except the intentional fixture finding, `mix dialyzer` 17 errors / 17 skipped / 1 unnecessary skip — byte-identical to `main`.

But there is one blocking problem, plus a set of message/errno mismatches I checked against GNU coreutils 9.11 and a measurable perf regression.

---

### Blocking: writes through a symlinked directory become unreadable at the path they were written to

`resolve_for_write`/`resolve_for_create` resolve intermediate symlinks; the read side does not. So the write lands at the resolved target and the caller can no longer see it at the path they used. Verified against `main` (`56c74b8`) — every one of these **worked before this PR** and works in real bash:

```
mkdir -p /real && ln -s /real /link

echo hi > /link/a.md   ; cat /link/a.md      # PR: exit 1 ENOENT  | main: "hi"  | bash: "hi"
echo a  >> /link/ap.txt; cat /link/ap.txt    # PR: exit 1          | main: ok    | bash: ok
cp /src.txt /link/c.txt; cat /link/c.txt     # PR: exit 1          | main: ok    | bash: ok
touch /link/t.txt      ; ls -l /link/t.txt   # PR: exit 1          | main: ok    | bash: ok
mkdir -p /link/sub     ; ls /link/sub        # PR: exit 1          | main: ok    | bash: ok
echo hi > /link/f      ; rm /link/f          # PR: exit 1          | main: ok    | bash: ok
cd /link; echo hi > f.txt; cat f.txt         # PR: exit 1          | main: "hi"  | bash: "hi"
```

This is the *same class of bug as #53*, mirrored: instead of "readable by path but not enumerable", it is now "written successfully but not readable at that path". Arguably worse for a caller, because the write reports success and silently places the bytes somewhere else.

The suite stays green because the one symlinked-directory test (`not_a_directory_test.exs:714`) reads back through the **target** (`/real/sub/a.md`), never through the link. A test that does `FS.write_file(fs, "/link/sub/a.md", ...)` then `FS.read_file(fs, "/link/sub/a.md")` fails today.

I think this has to land with the read side, not after it: extend the same component-wise walk to `stat/2`, `readdir/2`, `stream_read/3`, `exists?/2`, `rm/3` and `Memory.lstat/2`/`readlink/2` (a `resolve_for_read` that follows everything, and `resolve_for_create`-style ancestor-only resolution for `lstat`/`readlink`/`rm`, which must not follow the final link). Alternatively, hold the intermediate-symlink-following part back and ship only the ENOTDIR enforcement, which does not depend on it.

Fixing the read side also closes M4 and M3 below, and unblocks #57's `cp` from reporting GNU's `cannot stat` wording.

---

### Message / errno mismatches vs GNU coreutils 9.11

The ones that **match GNU exactly**: `touch`, `tee` (shallow), `ln -s`, `ln`, `mkdir` without `-p`, and `mkdir -p PATH` on an existing regular file (`File exists`). Nice work on those. The rest are inline: `mkdir -p` names the wrong path, `cp`/`mv` use the wrong verb for ENOTDIR, `tee` reports the wrong errno for a deeper path.

---

### Performance regression

Measured on this branch vs `56c74b8` (macOS, OTP 28, same machine, warm):

| benchmark | main | this PR | |
|---|---|---|---|
| 10k `FS.write_file` at depth 3 | 27 ms | 46 ms | 1.7x |
| 10k `FS.write_file` at depth 12 | 35 ms | 187 ms | 5.4x |
| 10k `FS.append_file` at depth 6 | 18 ms | 59 ms | 3.3x |
| `mkdir -p` depth 100 | 0.5 ms | 12 ms | 22x |
| `mkdir -p` depth 200 | 1.8 ms | 86 ms | 47x |
| `mkdir -p` depth 400 | 7.2 ms | 698 ms | **97x** |

O(depth) resolution per write is the intended cost, but it is O(depth) *string rebuilds* (`dirname` + `basename` + `join_child` at every level), so per-op cost is O(depth^2) bytes — that is the depth-12 column. The `mkdir -p` blowup is a separate, easily fixed issue; see the inline note on `do_mkdir_resolved`.

---

### Things I checked that are correct

- **No false ELOOP.** `seen` is threaded per chain and not shared across sibling ancestor resolutions, so a legitimate repeat of the same link in one path resolves: with `/L -> /real` and `/real/L2 -> /real`, `write_file("/L/L2/x")` correctly lands at `/real/x`. True cycles (`/a -> /b -> /a`, `/s -> /s`) return `:eloop` at every entry point (`write_file`, `append_file`, `mkdir -p`, `chmod`).
- **Root, `.`, relative symlink targets, dangling final component, cross-mount** all behave. `FS.new() |> VFS.mount("/mnt", Memory.new(%{"/j" => "f"}))` then `write_file("/mnt/j/a.md")` gives `{:enotdir, "/mnt/j/a.md", mount: "/mnt"}`.
- **`validate_initial_files!/1` — no false positives found.** It is genuinely component-wise: `%{"/m/jj" => _, "/m/j" => _}` passes; relative keys (`"m/j"`), `..` keys (`"/m/x/../j/a"`), trailing slashes (`"/m/j/"`) and extended-form values are all normalized and caught. Raising here is the right call, and the error message is genuinely good.
- **`resolve_for_create` for `mkdir`**: `mkdir -p /link` where `/link -> /real` (a directory) exits 0, matching GNU; `mkdir /link` gives `File exists`. The new `handle_existing_path/5` is what makes that work, and it is a real fix — `main` reported success for `mkdir -p` on a regular file.

### Also worth knowing (pre-existing, not introduced here)

- **`..` bypasses the new check.** `normalize/1` collapses `..` lexically before resolution, so `echo hi > /m/j/../k.md` still exits 0 and creates `/m/k.md`, where bash says `Not a directory` (and `gmkdir -p j/../k` says `cannot create directory 'j': Not a directory`). Not a regression, but the new docstrings claim "POSIX path resolution" and "resolves one component at a time", which this contradicts — worth one sentence of hedging in the moduledoc or UPGRADING item 7.
- **A failed redirect leaks the command's stdout.** `echo hi > /m/j/a.md` gives exit 1 and the right stderr, but `stdout == "hi\n"`; real bash produces no stdout because the redirect fails before the command runs. Same on `main` for `echo hi > /somedir`, so this PR did not cause it — but the PR turns a large class of previously-successful writes into failures, so it becomes reachable far more often, and in a sandbox that means file payloads landing in the caller's stdout. See the inline note on the redirect tests.
- **`symlink/3`/`link/3` still create missing parents** (`resolve_for_create` permits missing ancestors, then `ensure_parent_dirs/2` conjures them), so `ln -s /src /missing/dir/link` succeeds where bash gives ENOENT. Consistent with the documented `>` deviation; just noting it was not tightened.
- **`link/3` resolves the new path component-wise but leaves `existing_path` lexical** (memory.ex:239-243), so `ln /link/sub/f /dst` fails ENOENT. Another face of the blocking finding.

---

### Interaction with #57 (not a conflict to fix here, just for whoever merges second)

No semantic conflicts — the two approaches agree:

- **`cp.ex`**: #57 replaces `execute/3` wholesale, and its `file_result/3` fallback emits the identical `cp: cannot create regular file '#{dest}': #{strerror}` string. So this PR's `write_dest/4` becomes redundant on merge but the assertion in `not_a_directory_test.exs:865` survives unchanged. Take #57's version and drop `write_dest/4` — including its `:eisdir` comment, which #57 handles properly by copying *into* the directory rather than erroring.
- **`mv.ex`**: this PR collapses the `:enotdir` clause into a `%VFS.Error{}` catch-all; #57 keeps `:enotdir`, adds `:einval`, and adds its own catch-all. #57's clause set is a strict superset and produces the same `:enotdir` text, so take #57's.
- **`fs.ex`**: different regions (the `validate_initial_files!` defdelegate vs. the cp/mv self-copy guards). Textual only.
- One real interaction: #57's `dest_kind/2` stats the destination to decide "copy into directory", and with this PR that stat still returns `:enoent` for a path under a regular file, so neither PR can produce GNU's `cp: cannot stat 'j/a.md': Not a directory`. Fixing the read side above lets both land.
- #57's new `test/fixtures/bash_cases/cp.json` contains no ENOTDIR-through-a-file case, so no fixture conflict.

---

Summary: the design is right and the ENOTDIR half is solid and well tested. The symlinked-directory half needs its read-side counterpart before this can merge without regressing behavior that works today.

lib/just_bash/fs/memory.ex
davydog187: Fixed in c868a1e — took the read-side option rather than holding the symlink half back.

Resolution is now one walk shared by both sides. `resolve_path/2` follows every component including a symlink at the final one (reads, and `O_TRUNC`/`O_APPEND`/`chmod` writes); `resolve_for_create/2` resolves ancestors only, for the operations where the final name is the subject — `mkdir`/`symlink`/`link` (EEXIST rather than a link to follow) and `lstat`/`readlink`/`rm`. `stat`, `readdir`, `stream_read`, `exists?`, `rm`, `lstat`, `readlink`, and `link`'s `existing_path` all route through it.

Every row of your table now passes; there are tests for each, including `FS.write_file("/link/sub/a.md")` read back through `/link/sub/a.md` rather than the target.

Two notes on how it differs from the sketch:

**Symlink hops are budgeted, not remembered.** A single shared `seen` set across one resolution would have broken a case your old per-chain threading got right: with `/real/self -> /real`, `/real/self/self/x` is two hops to `/real/x`, and a set reports ELOOP on the second visit. So it is a hop counter capped at 40, matching Linux `SYMLOOP_MAX`. True cycles still give `:eloop` at every entry point, and there is a test pinning both the legitimate repeat and the genuine cycle.

**`exists?` resolves ancestors only.** Following the final component would have changed dangling-symlink semantics (`Map.has_key?` on the literal name is currently true for a dangling link), which is orthogonal to this bug.

This also let `do_resolve_entry/3` go — `__resolve_entry__/2` is built on the shared walk now — which removes every `@dialyzer :nowarn_function` in the file. See the reply on your nit thread.

lib/just_bash/fs/memory.ex
davydog187: Better than narrowed — the attribute is gone entirely.

Rebuilding resolution as a single pass removed the mutual recursion the skips existed for, and folding `__resolve_entry__/2` onto the same walk deleted `do_resolve_entry/3` and its separate skip. `memory.ex` now carries no `@dialyzer` attribute at all, and `mix dialyzer` still reports `Total errors: 17, Skipped: 17, Unnecessary Skips: 1` — identical to `main`.

lib/just_bash/fs/memory.ex
davydog187: Both fixed. `mkdir_with_parent/3` now calls `do_mkdir_resolved/3` directly instead of recursing through the protocol, and `resolve_full/3` is replaced by a single pass that splits the path into components once and descends with the resolved prefix as an accumulator.

Re-measured against a tree without this work, same machine:

| benchmark | before | as submitted | now |
|---|---|---|---|
| 10k `write_file` depth 3 | 19.5 ms | 46 ms | 24.9 ms |
| 10k `write_file` depth 12 | 33.8 ms | 187 ms | 48.3 ms |
| 10k `append_file` depth 6 | 29.9 ms | 59 ms | 39.2 ms |
| `mkdir -p` depth 100 | 0.5 ms | 12 ms | 0.3 ms |
| `mkdir -p` depth 200 | 1.6 ms | 86 ms | 0.8 ms |
| `mkdir -p` depth 400 | 6.6 ms | 698 ms | 4.2 ms |
| 10k `read_file` depth 12 | 33.4 ms | — | 34.3 ms |

`mkdir -p` is now faster than before the PR. The remaining ~1.3-1.4x on writes is the O(depth) component resolution you called the intended cost; the O(depth^2) *rebuilds* are gone. Reads pay ~3% despite now doing full resolution where they previously did a single map hit.

lib/just_bash/commands/mkdir.ex
davydog187: Both fixed.

**1.** `mkdir -p` names the offending ancestor now. Since `VFS.Error` has no field that survives `put_path/2`, the ancestor is found in `mkdir.ex` by walking the operand's own components and stopping at the first that stats as a non-directory — which also gets the display form right for free:

```
mkdir /m/j/2026           -> mkdir: cannot create directory '/m/j/2026': Not a directory
mkdir -p /m/j/2026        -> mkdir: cannot create directory '/m/j': Not a directory
mkdir -p /m/j/2026/deeper -> mkdir: cannot create directory '/m/j': Not a directory
cd /m && mkdir -p j/2026  -> mkdir: cannot create directory 'j': Not a directory
```

The test that pinned the old string is updated, and there is a new one for the relative form.

**2.** `handle_existing_path/5` falls back to `FS.lstat` when `FS.stat` fails, so a name that exists as a dangling link reports `File exists`:

```
ln -s /nowhere /dang; mkdir -p /dang
  -> mkdir: cannot create directory '/dang': File exists
```

Only a name absent from `lstat` too gets the stat error. `mkdir -p /link` where `/link -> /real` still exits 0.

lib/just_bash/commands/cp.ex
davydog187: Fixed, and you were right that it needed the read side first — with component-wise resolution `FS.write_file` returns `:enotdir` here, so `cp` can tell the two apart:

```
cp /src.txt /m/j/a.md  -> cp: cannot stat '/m/j/a.md': Not a directory
```

`:enotdir` gets `cannot stat`; every other error keeps `cannot create regular file`, with a test pinning the `:eisdir` case so the two do not collapse. The assertion at `not_a_directory_test.exs:865` is updated.

Left `mv`'s `cannot move 'X' to 'Y'` alone deliberately. It is pre-existing, and per your merge notes #57 keeps a clause set that produces that text — changing it here would make whichever lands second emit the wrong string. Worth a follow-up once #57 is in.

lib/just_bash/commands/tee.ex
davydog187: Resolved for free by the read side, exactly as you predicted — `tee.ex` is unchanged:

```
echo hi | tee /m/j/a.md       -> tee: /m/j/a.md: Not a directory
echo hi | tee /m/j/2026/a.md  -> tee: /m/j/2026/a.md: Not a directory
```

Added the deeper-path test that was missing, so it is no longer invisible in the suite.

test/just_bash/not_a_directory_test.exs
davydog187: Added — but the assertion fails as written, so the leak is fixed rather than tracked.

In `redirection.ex` the failure branches now clear the redirected stream before appending the shell's error. The bytes were bound for the file, so they must not surface as the caller's stdout:

```elixir
JustBash.exec(bash_with_file_parent(), "echo hi > /m/j/a.md")
#=> exit_code: 1, stderr: "bash: /m/j/a.md: Not a directory\n", stdout: ""
```

One helper covers all four operators and both streams — for `2>` it means the command's own stderr went to the file and only `bash: ...` survives, which is what bash does. All four tests assert `stdout`, plus a new one for the pre-existing `:eisdir` path you mentioned (`echo hi >> /d`).

Not a full fix: bash opens the target *before* running the command, so it produces no output because nothing ran. Here the command has already run, so its side effects still happen. Pre-flighting redirects is an executor change I kept out of this PR; the comment on `redirect_failed/4` says so.

lib/just_bash/fs/fs.ex
davydog187: Fixed the code rather than softening the doc. `Memory.new/1`'s reduce goes through a `seed!/4` helper that turns a write failure into `ArgumentError`, so all three entry points agree:

```elixir
iex> JustBash.FS.Memory.new(%{"/" => "x"})
** (ArgumentError) invalid initial files: cannot create "/": Is a directory

iex> JustBash.FS.new(%{"/" => "x"})
** (ArgumentError) invalid initial files: cannot create "/": Is a directory

iex> JustBash.new(files: %{"/" => "x"})
** (ArgumentError) invalid initial files: cannot create "/": Is a directory
```

Same wording as `init_filesystem/1` already produced. Both docstrings now name the two distinct failures — one path running through another, and a collision with a directory the backend already holds — with a test per entry point.

davydog187:
Thanks — the read-side diagnosis was right, and taking that option rather than holding the symlink half back turned out to close three of the other findings too. All eight threads are addressed in c868a1e and replied to individually; summary here.

## The blocking finding

Resolution is now **one walk shared by both sides**, split into components once and descending with the resolved prefix as an accumulator:

- `resolve_path/2` — follows every component including a symlink at the final one (reads, and `O_TRUNC`/`O_APPEND`/`chmod` writes)
- `resolve_for_create/2` — ancestors only, for the operations where the final name is the subject: `mkdir`/`symlink`/`link`, and `lstat`/`readlink`/`rm`

`stat`, `readdir`, `stream_read`, `exists?`, `rm`, `lstat`, `readlink`, and `link`'s `existing_path` all route through it. Every row of your regression table passes, with a test each — including the `FS.write_file("/link/sub/a.md")` → `FS.read_file("/link/sub/a.md")` case you noted the suite was missing.

Two deliberate deviations from the sketch:

- **Hops are budgeted (SYMLOOP_MAX 40), not remembered.** A shared `seen` set would reject a case your per-chain threading got right: with `/real/self -> /real`, `/real/self/self/x` is two hops to `/real/x`. Tests pin both that and genuine cycles.
- **`exists?` resolves ancestors only** — following the final component would change dangling-symlink semantics, which is orthogonal.

Side effect: `do_resolve_entry/3` is gone and `memory.ex` now carries **no `@dialyzer` attribute at all**, which subsumes the nit thread.

## Perf

`mkdir_with_parent/3` no longer recurses through the protocol, and the O(depth²) rebuilds are gone:

| benchmark | before | as submitted | now |
|---|---|---|---|
| 10k `write_file` depth 3 | 19.5 ms | 46 ms | 24.9 ms |
| 10k `write_file` depth 12 | 33.8 ms | 187 ms | 48.3 ms |
| 10k `append_file` depth 6 | 29.9 ms | 59 ms | 39.2 ms |
| `mkdir -p` depth 100 | 0.5 ms | 12 ms | 0.3 ms |
| `mkdir -p` depth 200 | 1.6 ms | 86 ms | 0.8 ms |
| `mkdir -p` depth 400 | 6.6 ms | 698 ms | 4.2 ms |
| 10k `read_file` depth 12 | 33.4 ms | — | 34.3 ms |

`mkdir -p` is now faster than before the PR. Reads pay ~3% for full resolution.

## Messages

All verified against your GNU 9.11 expectations:

```
mkdir -p /m/j/2026        -> mkdir: cannot create directory '/m/j': Not a directory
cd /m && mkdir -p j/2026  -> mkdir: cannot create directory 'j': Not a directory
ln -s /nowhere /dang; mkdir -p /dang
                          -> mkdir: cannot create directory '/dang': File exists
cp /src.txt /m/j/a.md     -> cp: cannot stat '/m/j/a.md': Not a directory
tee /m/j/2026/a.md        -> tee: /m/j/2026/a.md: Not a directory
```

## The stdout leak

Fixed, not just tracked — `assert result.stdout == ""` fails as written, so documenting it was not an option. The failure branches clear the redirected stream before appending the shell's error. Still not a full fix: bash opens the target before running the command, so side effects diverge; pre-flighting redirects is an executor change I kept out of scope and commented as such.

## Left alone deliberately

- **`mv`'s wording** — pre-existing, and per your merge notes #57 keeps a clause set producing that text. Changing it here would make whichever lands second emit the wrong string. Follow-up once #57 is in.
- **`symlink`/`link` creating missing parents** — you flagged it as not-tightened rather than wrong.
- **`..` collapsing lexically** — now stated as a known deviation in UPGRADING item 7 rather than hedged, since it does not admit unreachable state.

Also fixed the `link/3` lexical `existing_path` you spotted as another face of the blocking finding.

Five checks: 3,806 tests / 0 failures, `dialyzer` 17 errors / 17 skipped / 1 unnecessary skip (identical to `main`), `credo --strict` clean but for the intentional fixture finding, `format --check-formatted`, `compile --warnings-as-errors`.

davydog187:
Opened #59 for the remaining half of the redirect finding, and pointed the `redirect_failed/4` comment at it in 87a12c4.

To be precise about what is and is not left, since more of this was already correct than the review round suggested — verified on `c868a1e`:

| case | branch | bash |
|---|---|---|
| `echo hi > /adir` (`:eisdir`) | stdout `""` | `""` |
| `echo hi >> /adir`, `echo hi &> /adir` | stdout `""` | `""` |
| `ls / 2> /adir` | stdout = `ls` output | same (only stderr was redirected) |
| `echo hi > /m/j/a.md; echo code=$?` | `code=1` | `code=1` |
| `echo hi > /m/j/a.md && echo AND \|\| echo OR` | `OR` | `OR` |
| `false > f.txt` truncates `f.txt` | yes | yes |

`$?` propagation, `&&`/`||` short-circuiting, stdout suppression and truncate-on-open all match bash. Truncation matches because writing `""` is equivalent to opening with `O_TRUNC`.

The one divergence left is that the command body still executes, so its side effects survive: `mkdir /made > /m/j/x` creates `/made` where bash does not. Fixing that is not local — redirect target expansion has to move ahead of the body (targets can contain command substitution, so they cannot be expanded twice), and six call sites need the pre-flight: `do_execute_simple_command/6`, `For`/`While`/`Until`, `Subshell`, `Group`. Loops also open once for the whole loop in bash, which is its own decision. That is all-redirections surface with no bearing on ENOTDIR, so it is #59 rather than this PR.

ivarvong:
Three things to fix at `c868a1e` — all from the same cause: now that `stat`/`readdir` resolve symlinks, traversal descends *through* them and two commands see error kinds they don't handle. Everything else checks out here: gates reproduce (3806 tests, dialyzer 17/17 same as main), the hop budget matches Linux exactly, and `mkdir -p` at depth 400 is genuinely faster than main (4.2 ms vs 8.6 ms).

**1. Recursive traversal doesn't terminate through a symlinked directory.**

```elixir
b = JustBash.new(cwd: "/w")
JustBash.exec(b, "mkdir -p /w/real; echo hi > /w/real/a.txt; ln -s /w/real /w/real/self; find /w/real | wc -l")
#=> "82\n"        main: "3\n"        GNU find: 3

JustBash.exec(b, "mkdir -p /w/real; ln -s /w/real /w/real/s1; ln -s /w/real /w/real/s2; find /w/real")
# never returns (2^40 paths, bounded only by @symloop_max) — same for du -s, grep -r, tree, and FS.walk
```

Main returns instantly for both. GNU doesn't descend into symlinks at all under the default `-P`, so the one-link case is wrong output before it's a hang. Two `ln -s` calls from an untrusted script wedge the exec, and dol walks each mount root after every command (#53), so it'd wedge there.

Two spots to fix it: commands should decide descent with `lstat` rather than `stat` (that's `-P`, and it fixes the 82-vs-3 too), and `FS.walk` wants a native `walk/3` in the `Memory` defimpl — with ENOTDIR enforced the map's keys are exactly the reachable tree, so prefix enumeration is exact, O(n), and never follows a link.

**2. `cd` through a regular file raises out of `JustBash.exec/2`.**

```elixir
JustBash.exec(JustBash.new(files: %{"/m/j" => "x"}), "cd /m/j/sub")
# ** (CaseClauseError) {:error, %VFS.Error{kind: :enotdir}} at cd.ex:22
# main: exit 1, "bash: cd: /m/j/sub: No such file or directory"
```

The `case` at `cd.ex:22` has `:ok`, `:ok`, `:enoent` and no fallback. Same clause also covers `cd` into a link loop, which raises on main too.

**3. `ls` on a symlink loop raises.**

```elixir
JustBash.exec(JustBash.new(), "ln -s /b /a; ln -s /a /b; ls /a")
# ** (CaseClauseError) {:error, %VFS.Error{kind: :eloop}} at ls.ex:35
# main: exit 1. GNU: exit 2, "Too many levels of symbolic links"
```

`list_path/4` handles `:enoent` (`ls.ex:40`) and `:enotdir` (`ls.ex:43`) only.

The 66 tests are thorough on the ENOTDIR side, but nothing walks a tree that *contains* a link cycle — a "walk terminates" invariant next to the existing "walk reaches everything" one would catch #1.

A few smaller non-blocking things (a `2>` stdout leak, trailing-slash handling, the `%{content:, mode:}` form through `JustBash.new/1`) — I'll file those as issues rather than pile on here.

davydog187:
All three fixed at `9bf1bb1`, plus the invariant you asked for. One shared cause, as you diagnosed: `stat/2` now resolves symlinks, so every "should I descend?" question answered through it descended.

**1. Recursive traversal.** Both spots you named:

*Commands decide descent with `lstat`.* `find`, `du`, `tree`, and `grep -r` treat a symlink as a leaf whatever it points at — GNU's default `-P`. That fixes the 82-vs-3 as the same edit, since `find` no longer walks through the link:

```elixir
b = JustBash.new(cwd: "/w")
JustBash.exec(b, "mkdir -p /w/real; echo hi > /w/real/a.txt; ln -s /w/real /w/real/self; find /w/real | wc -l")
#=> "3\n"                                                    GNU find: 3

JustBash.exec(b, "mkdir -p /w/real; ln -s /w/real /w/real/s1; ln -s /w/real /w/real/s2; find /w/real")
#=> "/w/real\n/w/real/s1\n/w/real/s2\n"                       instant
```

Two knock-ons worth flagging: `-type f` and `-type d` no longer match symlinks, so `-type l` is now accepted and does. And `grep -r` skips links met while recursing but still follows one named as an operand — the split GNU documents. `du` counts a symlink's own size and none of its target's.

*`Memory` gets a native `walk/3`.* Exactly your reasoning: with ENOTDIR enforced the keys are the reachable tree, so one prefix scan builds a `dir => sorted children` index and the depth-first walk over it is exact and O(n). A symlink is an entry, never an edge, so it cannot diverge. The composed walk's contract is preserved — laziness (every stage is a `Stream`, so `Stream.take/2` still halts early), `:max_depth`, `:include_dirs`, and a symlinked root still being followed with emitted paths keeping the caller's spelling of it. The one visible change is that a symlink now reports `type: :symlink` rather than its target's type, which is the point.

`cp -r`, `rm -rf`, `ls -R`, globbing, and `wc` over the cyclic tree all terminate too — checked, no other traversal was deciding descent through `stat`.

**2. `cd`** and **3. `ls`** both render `FS.strerror/1` now instead of enumerating kinds, so every kind reports rather than raising:

```
cd /m/j/sub   → bash: cd: /m/j/sub: Not a directory                          exit 1
cd /a         → bash: cd: /a: Too many levels of symbolic links              exit 1
ls /a         → ls: cannot access '/a': Too many levels of symbolic links    exit 1
```

The `cd` wording is checked against real bash, which the previous ENOENT text did not match either. `ls` stays exit 1 rather than GNU's 2 — the module's existing ENOENT path is also 1, and making just the new kind disagree seemed worse than the shared deviation.

**Tests.** 19 added, your repros among them. The one that would have caught this is `walk/3 terminates over a tree that contains a symlink cycle`, sitting next to the existing reaches-everything test, plus per-command termination cases. Divergent runs are bounded by a task with a 5 s timeout and `:brutal_kill`, so a regression fails the test instead of hanging the suite (or OOM-ing it).

Gates: 3,825 tests / 0 failures, dialyzer 17/17 unchanged, credo and format clean, `mix docs` warning count unchanged. `mkdir -p` at depth 400 is untouched by this round — 23.2 ms before, 23.3 ms after on my machine, so I could not reproduce your 4.2 ms absolute number, likely harness differences; the before/after comparison is what matters here.

`UPGRADING.md` item 7 gains the traversal delta. One thing I deliberately did not do: `tree` lists a symlink as a leaf but does not render `name -> target` the way real `tree` does. That arrow is a pre-existing gap rather than something this fix should invent — happy to take it as a follow-up issue alongside the smaller ones you're filing.

ivarvong (approved):

PR #57, opened 2026-07-29
fix: cp into a directory, and make cp -r actually copy

### LINKED ISSUE #52: cp into an existing directory raises MatchError; cp -r is a no-op
Two independent `cp` bugs, both present in hex **0.3.0** and still on **main** (`56c74b8`). Both are ordinary bash usage, and `mv` gets the same case right — which points at the fix.

Verified against real bash (GNU/BSD `cp`, macOS) and against both versions.

## 1. `cp SRC DIR` raises an uncaught `MatchError`

```elixir
b = JustBash.new(files: %{"/m/a.md" => "A\n", "/m/d/k.md" => "K\n"})
JustBash.exec(b, "cp /m/a.md /m/d/")
```

```
** (MatchError) no match of right hand side value:

    {:error, %VFS.Error{kind: :eisdir, path: "/m/d", mount: "/", message: ":eisdir at /m/d"}}
```

On 0.3.0 the same call raises with `{:error, :eisdir}`. Both the trailing-slash and bare forms raise:

| command | real bash | just_bash |
|---|---|---|
| `cp /m/a.md /m/d/` | exit 0, creates `/m/d/a.md` | **raises `MatchError`** |
| `cp /m/a.md /m/d` | exit 0, creates `/m/d/a.md` | **raises `MatchError`** |
| `mv /m/a.md /m/d/` | exit 0 | exit 0 ✅ |

The raise escapes `JustBash.exec/2`, so a host embedding the sandbox gets an exception rather than a non-zero result — and one that names an Elixir internal, so it can't be shown to whatever drove the command.

### Suggested fix

`Cp.execute/3` hard-matches `{:ok, fs} = FS.write_file(...)`, which answers `{:error, :eisdir}` when the destination is an existing directory. `Mv.execute/3` already stats the destination and appends the basename in that case; `cp` should do the same. The 0.4 filesystem rewrite made cp/mv "compositions over vfs primitives" but the hard match survived — only the error struct changed.

## 2. `cp -r SRC DIR` copies nothing and reports a missing operand

```elixir
b = JustBash.new(files: %{"/m/s/x.md" => "X\n", "/m/d/k.md" => "K\n"})
{r, _} = JustBash.exec(b, "cp -r /m/s /m/d")
# r.exit_code #=> 1
# r.stderr    #=> "cp: missing file operand\n"
```

Two operands are present. Real bash exits 0 and creates `d/s/x.md`. Every form fails identically, so it isn't about the trailing slash or whether the destination exists:

```
cp -r /m/s /m/d        exit=1  "cp: missing file operand\n"   nothing copied
cp -R /m/s /m/d        exit=1  "cp: missing file operand\n"   nothing copied
cp -r /m/s/ /m/d/      exit=1  "cp: missing file operand\n"   nothing copied
cp -r /m/s /m/newdir   exit=1  "cp: missing file operand\n"   nothing copied
```

Reads like flag parsing consuming an operand before the arity check. Recursive copy appears to be unreachable through the shell at present.

## Impact

`cp` is a top-tier command for an agent moving files around a sandbox: #1 crashes the host on a normal command, and #2 makes `cp -r` silently useless while looking like user error. Worth noting #1 is only survivable downstream by wrapping `JustBash.exec/2` in a `rescue`, which is what [dol](https://github.com/davydog187/dol) had to do.

davydog187:
Closes #52

Two independent `cp` bugs, plus one hang they exposed.

## Bug 1: `cp SRC DIR` raised an uncaught MatchError

**Root cause.** `Cp.execute/3` never looked at the destination: it read the source and wrote the content straight to the destination path with a hard match, `{:ok, fs} = FS.write_file(fs, dest_resolved, content)`. When the destination was an existing directory, `write_file` returned `{:error, %VFS.Error{kind: :eisdir}}`, the match failed, and the `MatchError` escaped `JustBash.exec/2` and crashed the host. `mv` already did the right thing (stat the destination, append the source basename when it is a directory); `cp` did not.

**Fix.** `cp` stats the destination and, when it is a directory, copies into it under the source's basename — for `cp SRC DIR` and `cp SRC DIR/` alike. No filesystem result is hard-matched any more; each error kind becomes a bash-shaped message on stderr with exit 1 (same precedent as dffd5dd for `mkdir`/`rm`).

## Bug 2: `cp -r SRC DIR` copied nothing and reported "missing file operand"

**Root cause.** `cp` parsed no flags at all. `-r` was therefore just another operand, `["-r", "/m/s", "/m/d"]` did not match the `[src, dest]` clause, and the catch-all arity error fired. Recursive copy was unreachable through the shell even though `FS.cp/4` has supported `recursive: true` all along.

**Fix.** `cp` parses flags through the shared `JustBash.FlagParser`: `-r`, `-R`, `--recursive`, and `-a`/`--archive` (implies recursive) enable directory copies; `-f`/`--force` and `-p`/`--preserve` are accepted no-ops (a copy always overwrites, and always carries the source's mode and mtime across). Combined forms (`-rf`) and `--` work through the parser. Directory copies go through `FS.cp(recursive: true)`, so a source directory copies *into* an existing destination directory (`d/s/...`) and *as* the destination when it does not exist — matching bash.

## Also fixed: infinite recursion on a copy into its own subtree

Enabling `-r` made a latent `FS.cp/4` hang reachable: copying a directory into a path inside itself recursed forever, because each pass created new children under the tree still being walked. It was already reachable today through `mv /a /a/b`, which hung the exec. `FS.cp/4` now rejects that up front with `:einval` before mutating anything, and the callers translate it to bash's wording:

- `cp: cannot copy a directory, '/a', into itself, '/a/b/a'`
- `mv: cannot move '/a' to a subdirectory of itself, '/a/b/a'`

`Mv` also gained a catch-all `%VFS.Error{}` clause, so a new error kind can no longer raise a `CaseClauseError` out of `mv`.

## Behaviors changed

- `cp SRC DIR` / `cp SRC DIR/` copy into the directory, exit 0 (was: MatchError crash).
- `cp -r`/`-R`/`--recursive`/`-a` copy directory trees, including merging into an existing destination subtree and the `cp -r src/. dest` contents idiom.
- Any number of sources may be copied into a trailing directory operand (`cp *.md dir/`). The trailing operand must be an existing directory, else `cp: target 'x': Not a directory` / `No such file or directory`, as in bash. A failing source no longer stops the rest: bash copies what it can, reports each failure, exits 1.
- Errors now match GNU coreutils wording: `missing file operand` / `missing destination file operand after 'x'` (both with the `Try 'cp --help'` line), `cannot stat`, `-r not specified; omitting directory`, `cannot overwrite non-directory 'd' with directory 's'`, `'a' and 'a' are the same file`, `target 'c': Not a directory`.
- A symlink source is dereferenced (bash's default without `-d`/`-P`); a regular-file copy now carries the source's mode across (it previously landed as 0644).
- Unchanged divergence, shared with `mv`: this filesystem creates missing destination parent directories instead of failing with ENOENT. Covered by a test with a comment so it is deliberate rather than accidental.

## Tests

Written first, all failing before the fix (34 failures, including the `mv` recursion timeout).

- `test/commands/file_operations_test.exs`: 30 new `cp` cases (directory destinations, every recursive form, trailing slashes, `dir/.`, merges, nested trees, multi-source, mode, symlinks, `--`, and each error message) plus the `mv` subdirectory-of-itself case.
- `test/just_bash/fs_test.exs`: `FS.cp/4` merge and self-copy guards, `FS.mv/3` subdirectory guard, and a sibling-prefix case (`/ab` -> `/abc`) so the guard cannot over-match on a shared path prefix.
- `test/fixtures/bash_cases/cp.json` + `bash_expected/cp.json`: 20 new bash comparison cases recorded from GNU coreutils via `mix bash_fixtures cp` (there were none for `cp`), asserting stdout, stderr, and exit code against real bash.

All five gates pass: `mix compile --warnings-as-errors`, `mix dialyzer` (17 errors / 17 skipped, unchanged from main), `mix credo --strict` (only the pre-existing intentional test-fixture finding), `mix format --check-formatted`, `mix test` (3798 tests, 0 failures; also green with `--include bash_comparison`).

### DIFF
--- a/lib/just_bash/commands/cp.ex
+++ b/lib/just_bash/commands/cp.ex
@@ -2,14 +2,25 @@ defmodule JustBash.Commands.Cp do
   @moduledoc """
   The `cp` command - copy files and directory trees.
 
-  Flags: `-r`/`-R`/`--recursive` and `-a`/`--archive` (which implies
-  `--recursive`) copy directories. `-f`/`--force` and `-p`/`--preserve`
-  are accepted and do nothing: a copy always overwrites its destination,
-  and always carries the source's mode and mtime across.
+  Flags:
 
-  As in bash, a destination that is an existing directory receives the
-  source under its own basename, and any number of sources may be copied
-  into a trailing directory operand.
+    * `-r`/`-R`/`--recursive`, and `-a`/`--archive` (which implies
+      `--recursive`) — copy directory trees.
+    * `-n`/`--no-clobber` — leave an existing destination alone. Bash decides
+      this per file inside a tree; here it is decided per operand.
+    * `-P`/`--no-dereference` (and `-d`) / `-L`/`--dereference` — copy a
+      command-line symlink as a link, or follow it. As in bash, a shallow copy
+      follows the link and a recursive copy preserves it; `-L` and `-P`
+      override that. Links *inside* a copied tree are always preserved.
+    * `-v`/`--verbose` — print `'src' -> 'dest'` for every copied entry.
+    * `-f`/`--force`, `-p`/`--preserve` and `-i`/`--interactive` — accepted,
+      with nothing to do: a copy always overwrites, always carries the
+      source's mode and mtime across, and a sandbox has no terminal to prompt
+      on (so `-i` proceeds as if the prompt were answered yes).
+
+  As in bash, a destination that is an existing directory receives the source
+  under its own basename, and any number of sources may be copied into a
+  trailing directory operand.
   """
   @behaviour JustBash.Commands.Command
 
@@ -17,19 +28,49 @@ defmodule JustBash.Commands.Cp do
   alias JustBash.FlagParser
   alias JustBash.FS
 
-  # `-R` and the long forms are aliases of `-r`; FlagParser sees `--recursive`
-  # as the flag string "-recursive".
+  defmodule Options do
+    @moduledoc false
+
+    @enforce_keys [:mode, :links, :clobber, :verbose?]
+    defstruct [:mode, :links, :clobber, :verbose?]
+
+    @type t :: %__MODULE__{
+            mode: :recursive | :shallow,
+            links: :follow | :preserve,
+            clobber: :always | :never,
+            verbose?: boolean()
+          }
+  end
+
+  # `-R` and the long forms are aliases of their short flags; FlagParser sees
+  # `--recursive` as the flag string "-recursive".
   @flag_spec %{
-    boolean: [:r, :a, :f, :p],
+    boolean: [:r, :a, :f, :p, :n, :i, :v, :P, :L, :d],
     aliases: %{
       "R" => :r,
       "-recursive" => :r,
       "-archive" => :a,
       "-force" => :f,
-      "-preserve" => :p
+      "-preserve" => :p,
+      "-no-clobber" => :n,
+      "-interactive" => :i,
+      "-verbose" => :v,
+      "-no-dereference" => :P,
+      "-dereference" => :L
     },
     value: [],
-    defaults: %{r: false, a: false, f: false, p: false}
+    defaults: %{
+      r: false,
+      a: false,
+      f: false,
+      p: false,
+      n: false,
+      i: false,
+      v: false,
+      P: false,
+      L: false,
+      d: false
+    }
   }
 
   @try_help "Try 'cp --help' for more information.\n"
@@ -40,102 +81,192 @@ defmodule JustBash.Commands.Cp do
   @impl true
   def execute(bash, args, _stdin) do
     {flags, operands} = FlagParser.parse(args, @flag_spec)
-    copy(bash, operands, flags.r or flags.a)
+    copy(bash, operands, options(flags))
   end
 
-  defp copy(bash, [], _recursive?),
+  defp options(flags) do
+    mode = if flags.r or flags.a, do: :recursive, else: :shallow
+
+    %Options{
+      mode: mode,
+      links: link_mode(flags, mode),
+      clobber: if(flags.n, do: :never, else: :always),
+      verbose?: flags.v
+    }
+  end
+
+  # Bash dereferences a command-line symlink for a shallow copy only: `-r` and
+  # `-a` imply `-d`. `-L` forces the link to be followed, `-P`/`-d` to be kept.
+  defp link_mode(%{L: true}, _mode), do: :follow
+  defp link_mode(%{P: true}, _mode), do: :preserve
+  defp link_mode(%{d: true}, _mode), do: :preserve
+  defp link_mode(_flags, :recursive), do: :preserve
+  defp link_mode(_flags, :shallow), do: :follow
+
+  defp copy(bash, [], _opts),
     do: {Command.error("cp: missing file operand\n" <> @try_help), bash}
 
-  defp copy(bash, [src], _recursive?),
+  defp copy(bash, [src], _opts),
     do:
       {Command.error("cp: missing destination file operand after '#{src}'\n" <> @try_help), bash}
 
   # Two operands: the destination names the copy, unless it already exists as a
   # directory — then the copy lands inside it under the source's basename.
-  defp copy(bash, [src, dest], recursive?) do
+  defp copy(bash, [src, dest], opts) do
     case dest_kind(bash, dest) do
-      {:directory, bash} -> copy_one(bash, src, dest_in_dir(dest, src), recursive?)
-      {_kind, bash} -> copy_one(bash, src, dest, recursive?)
+      {:ok, :directory, bash} -> copy_one(bash, src, dest_in_dir(dest, src), opts)
+      {:ok, :other, bash} -> copy_one(bash, src, dest, opts)
+      {:error, _error, bash} -> copy_one(bash, src, dest, opts)
     end
   end
 
   # Three or more operands: the last one has to be an existing directory.
-  defp copy(bash, operands, recursive?) do
+  defp copy(bash, operands, opts) do
     {sources, [dest]} = Enum.split(operands, -1)
 
     case dest_kind(bash, dest) do
-      {:directory, bash} ->
-        copy_each(bash, sources, dest, recursive?)
+      {:ok, :directory, bash} ->
+        copy_each(bash, sources, dest, opts)
 
-      {{:error, error}, bash} ->
-        {Command.error("cp: target '#{dest}': #{FS.strerror(error)}\n"), bash}
-
-      {_kind, bash} ->
+      {:ok, :other, bash} ->
         {Command.error("cp: target '#{dest}': Not a directory\n"), bash}
+
+      {:error, error, bash} ->
+        {Command.error("cp: target '#{dest}': #{FS.strerror(error)}\n"), bash}
     end
   end
 
+  # An empty operand names no file. Resolving it would produce the cwd, which is
+  # not what bash does with it.
+  defp dest_kind(bash, ""), do: {:error, VFS.Error.new(:enoent, path: ""), bash}
+
   defp dest_kind(bash, dest) do
     case FS.stat(bash.fs, FS.resolve_path(bash.cwd, dest)) do
-      {:ok, %VFS.Stat{type: :directory}, fs} -> {:directory, %{bash | fs: fs}}
-      {:ok, %VFS.Stat{}, fs} -> {:other, %{bash | fs: fs}}
-      {:error, %VFS.Error{} = error} -> {{:error, error}, bash}
+      {:ok, %VFS.Stat{type: :directory}, fs} -> {:ok, :directory, %{bash | fs: fs}}
+      {:ok, %VFS.Stat{}, fs} -> {:ok, :other, %{bash | fs: fs}}
+      {:error, %VFS.Error{} = error} -> {:error, error, bash}
     end
   end
 
   # A failing source does not abort the run: bash copies what it can, reports
   # every failure, and exits 1.
-  defp copy_each(bash, sources, dest, recursive?) do
-    {stderr, exit_code, bash} =
-      Enum.reduce(sources, {[], 0, bash}, fn src, {err_acc, code_acc, bash_acc} ->
-        {result, bash_acc} = copy_one(bash_acc, src, dest_in_dir(dest, src), recursive?)
-        {[err_acc, result.stderr], max(code_acc, result.exit_code), bash_acc}
+  defp copy_each(bash, sources, dest, opts) do
+    {stdout, stderr, exit_code, bash} =
+      Enum.reduce(sources, {[], [], 0, bash}, fn src, {out_acc, err_acc, code_acc, bash_acc} ->
+        {result, bash_acc} = copy_one(bash_acc, src, dest_in_dir(dest, src), opts)
+
+        {[out_acc, result.stdout], [err_acc, result.stderr], max(code_acc, result.exit_code),
+         bash_acc}
       end)
 
-    {Command.result("", IO.iodata_to_binary(stderr), exit_code), bash}
+    {Command.result(IO.iodata_to_binary(stdout), IO.iodata_to_binary(stderr), exit_code), bash}
   end
 
-  defp copy_one(bash, src, dest, recursive?) do
+  defp copy_one(bash, "", _dest, _opts),
+    do: {Command.error("cp: cannot stat '': No such file or directory\n"), bash}
+
+  defp copy_one(bash, src, "", opts), do: empty_dest(bash, src, opts)
+
+  defp copy_one(bash, src, dest, opts) do
     src_path = FS.resolve_path(bash.cwd, src)
     dest_path = FS.resolve_path(bash.cwd, dest)
 
     if src_path == dest_path do
       {Command.error("cp: '#{src}' and '#{dest}' are the same file\n"), bash}
     else
-      copy_resolved(bash, {src, src_path}, {dest, dest_path}, recursive?)
+      keep_or_copy(bash, {src, src_path}, {dest, dest_path}, opts)
     end
   end
 
-  defp copy_resolved(bash, {src, src_path}, {dest, dest_path}, recursive?) do
+  # `-n` leaves an existing destination alone, and says nothing about it.
+  defp keep_or_copy(bash, src, {_dest, dest_path} = dest, %Options{clobber: :never} = opts) do
+    case FS.exists?(bash.fs, dest_path) do
+      {true, fs} -> {Command.ok(), %{bash | fs: fs}}
+      {false, fs} -> copy_resolved(%{bash | fs: fs}, src, dest, opts)
+    end
+  end
+
+  defp keep_or_copy(bash, src, dest, %Options{clobber: :always} = opts),
+    do: copy_resolved(bash, src, dest, opts)
+
+  defp copy_resolved(bash, {src, src_path}, {dest, dest_path}, opts) do
     case FS.lstat(bash.fs, src_path) do
       {:ok, %VFS.Stat{type: :directory}, fs} ->
-        copy_dir(%{bash | fs: fs}, {src, src_path}, {dest, dest_path}, recursive?)
+        copy_dir(%{bash | fs: fs}, {src, src_path}, {dest, dest_path}, opts)
 
-      # Without `-d`/`-P`, bash copies what a link points at, not the link.
       {:ok, %VFS.Stat{type: :symlink}, fs} ->
-        copy_dereferenced(%{bash | fs: fs}, {src, src_path}, {dest, dest_path})
+        copy_link(%{bash | fs: fs}, {src, src_path}, {dest, dest_path}, opts)
 
       {:ok, %VFS.Stat{}, fs} ->
         fs
         |> FS.cp(src_path, dest_path)
-        |> file_result(%{bash | fs: fs}, dest)
+        |> file_result(%{bash | fs: fs}, {src, src_path}, dest, opts)
 
       {:error, %VFS.Error{} = error} ->
         {Command.error("cp: cannot stat '#{src}': #{FS.strerror(error)}\n"), bash}
     end
   end
 
-  defp copy_dir(bash, {src, _src_path}, {_dest, _dest_path}, false),
+  # A preserved link is copied as a link — dangling or not, directory or not.
+  defp copy_link(bash, {src, src_path}, {dest, dest_path}, %Options{links: :preserve} = opts) do
+    bash.fs
+    |> FS.cp(src_path, dest_path)
+    |> file_result(bash, {src, src_path}, dest, opts)
+  end
+
+  # A followed link copies whatever it names, under the link's own name: a link
+  # to a directory therefore needs `-r`, exactly like a directory does.
+  defp copy_link(bash, {src, src_path}, dest, %Options{links: :follow} = opts) do
+    case link_target(bash.fs, src_path, 0) do
+      {:ok, target_path, fs} ->
+        copy_resolved(%{bash | fs: fs}, {src, target_path}, dest, opts)
+
+      {:error, %VFS.Error{} = error} ->
+        {Command.error("cp: cannot stat '#{src}': #{FS.strerror(error)}\n"), bash}
+    end
+  end
+
+  @max_link_hops 32
+
+  # The path a symlink chain ends at, so a followed copy can hand `FS.cp/4` a
+  # real file or directory instead of a link.
+  defp link_target(_fs, path, hops) when hops >= @max_link_hops,
+    do: {:error, VFS.Error.new(:eloop, path: path)}
+
+  defp link_target(fs, path, hops) do
+    case FS.lstat(fs, path) do
+      {:ok, %VFS.Stat{type: :symlink}, fs} -> follow_link(fs, path, hops)
+      {:ok, %VFS.Stat{}, fs} -> {:ok, path, fs}
+      {:error, %VFS.Error{} = error} -> {:error, error}
+    end
+  end
+
+  defp follow_link(fs, path, hops) do
+    case FS.readlink(fs, path) do
+      {:ok, target, fs} ->
+        link_target(fs, FS.resolve_path(FS.dirname(path), target), hops + 1)
+
+      {:error, %VFS.Error{} = error} ->
+        {:error, error}
+    end
+  end
+
+  defp copy_dir(bash, {src, _src_path}, {_dest, _dest_path}, %Options{mode: :shallow}),
     do: {Command.error("cp: -r not specified; omitting directory '#{src}'\n"), bash}
 
-  defp copy_dir(bash, {src, src_path}, {dest, dest_path}, true) do
+  defp copy_dir(bash, {src, src_path}, {dest, dest_path}, %Options{mode: :recursive} = opts) do
     case FS.cp(bash.fs, src_path, dest_path, recursive: true) do
       {:ok, fs} ->
-        {Command.ok(), %{bash | fs: fs}}
+        report(%{bash | fs: fs}, {src, src_path}, dest, opts)
 
       {:error, %VFS.Error{kind: :einval}} ->
         {Command.error("cp: cannot copy a directory, '#{src}', into itself, '#{dest}'\n"), bash}
 
+      # Verified against bash for a destination that already exists as a regular
+      # file. A destination whose *ancestor* is a regular file surfaces the same
+      # kind, and bash words that one "cannot stat '<dest>': Not a directory" —
+      # telling the two apart needs an lstat of the destination first, so read
+      # this wording as unverified for that case.
       {:error, %VFS.Error{kind: :enotdir}} ->
         {Command.error("cp: cannot overwrite non-directory '#{dest}' with directory '#{src}'\n"),
          bash}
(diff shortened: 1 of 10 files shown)


lib/just_bash/commands/cp.ex
```diff
@@ -1,31 +1,180 @@
...
+  end
+
+  defp copy_resolved(bash, {src, src_path}, {dest, dest_path}, recursive?) do
+    case FS.lstat(bash.fs, src_path) do
+      {:ok, %VFS.Stat{type: :directory}, fs} ->
+        copy_dir(%{bash | fs: fs}, {src, src_path}, {dest, dest_path}, recursive?)
+
+      # Without `-d`/`-P`, bash copies what a link points at, not the link.
+      {:ok, %VFS.Stat{type: :symlink}, fs} ->
+        copy_dereferenced(%{bash | fs: fs}, {src, src_path}, {dest, dest_path})
```
davydog187: **Medium: `cp -r` on a symlink-to-directory copies nothing, and the non-recursive message is wrong.**

This clause is reached before any directory check and always ends in `read_file`, so a link that points at a directory can never be copied. Verified on this branch:

```
ln -s /sd /sdlink          # /sd is a directory
cp    /sdlink /outdir      # rc=1  cp: cannot stat '/sdlink': Is a directory
cp -r /sdlink /outdir2     # rc=1  cp: cannot stat '/sdlink': Is a directory   (nothing created)
```

GNU coreutils 9.x, same setup:

```
cp    sdlink outdir        # rc=1  cp: -r not specified; omitting directory 'sdlink'
cp -r sdlink outdir2       # rc=0  outdir2 -> sd     (the link itself is copied)
```

Also worth noting the comment above: "Without `-d`/`-P`, bash copies what a link points at" holds only for a *non*-recursive copy. Under `-r` (and `-a`, which GNU defines to include `-d`) GNU preserves command-line symlinks — I confirmed both `cp -r alink rcopy` and `cp -a alink acopy` produce symlinks, while this branch dereferences them. `FS.cp/4` already does the right thing here (`cp_symlink`), so the recursive case can just delegate:

```elixir
{:ok, %VFS.Stat{type: :symlink}, fs} when recursive? ->
  fs |> FS.cp(src_path, dest_path) |> file_result(%{bash | fs: fs}, dest)

{:ok, %VFS.Stat{type: :symlink}, fs} ->
  # follow, then reuse copy_dir/copy_regular so a link to a directory
  # gets GNU's "-r not specified; omitting directory" message
  ...
```

lib/just_bash/commands/cp.ex
```diff
@@ -1,31 +1,180 @@
...
+      {:error, %VFS.Error{} = error} ->
+        {Command.error("cp: cannot create directory '#{dest}': #{FS.strerror(error)}\n"), bash}
+    end
+  end
+
+  defp copy_dereferenced(bash, {src, src_path}, {dest, dest_path}) do
+    case FS.read_file(bash.fs, src_path) do
+      {:ok, content, fs} ->
+        fs
+        |> FS.write_file(dest_path, content)
```
davydog187: **Medium/low: this path loses the source mode**, which contradicts the new moduledoc ("always carries the source's mode and mtime across"). `FS.write_file/3` with no opts lands 0644, while `FS.cp/4`'s `cp_regular` passes `mode:`/`mtime:`.

Verified on this branch: source written with `mode: 0o755`, copied *through a symlink* -> destination mode `0644`. GNU keeps `0755` (`cp mlink mcopy` -> `-rwxr-xr-x`). The plain-file path is fine (test at `file_operations_test.exs` covers it) — it's only the dereferenced-symlink path that regresses. Resolving the link and delegating to `FS.cp/4` on the target fixes finding #1 and this one together.

lib/just_bash/commands/cp.ex
```diff
@@ -1,31 +1,180 @@
...
+    do: {Command.error("cp: cannot create regular file '#{dest}': #{FS.strerror(error)}\n"), bash}
+
+  # The destination of a copy into a directory, spelled the way bash spells it
+  # in messages: the directory operand with the source's basename appended. The
+  # `dir/.` idiom copies the source's *contents*, so it keeps the destination.
+  defp dest_in_dir(dest, src) do
+    if dot_segment?(src) do
+      dest
+    else
+      String.trim_trailing(dest, "/") <> "/" <> FS.basename(src)
```
davydog187: **Low: an empty destination operand lands in the wrong directory and exits 0.** This composes the destination from the operand *spelling*, which is correct for every relative spelling I tried (`cd /work && cp a.md d` -> `/work/d/a.md`, good) — except `""`, which resolves to the cwd but concatenates into an absolute path:

```
cd /work && cp /x/a.md ''
# this branch: rc=0, writes /a.md (root), /work untouched
# GNU:         rc=1, cp: cannot create regular file '': No such file or directory
```

Cheap guard: reject an empty operand up front, or build from the resolved destination and only use the spelling for the message.

lib/just_bash/commands/cp.ex
```diff
@@ -1,31 +1,180 @@
...
   @behaviour JustBash.Commands.Command
 
   alias JustBash.Commands.Command
+  alias JustBash.FlagParser
   alias JustBash.FS
 
+  # `-R` and the long forms are aliases of `-r`; FlagParser sees `--recursive`
+  # as the flag string "-recursive".
+  @flag_spec %{
+    boolean: [:r, :a, :f, :p],
```
davydog187: **Low: GNU options outside this spec now parse as filenames.** `FlagParser` falls through to `:not_a_flag` for anything unknown, so every one of these becomes an operand:

```
cp -v /a.md /d        -> rc=1  cp: cannot stat '-v': No such file or directory
cp -n /a.md /d        -> rc=1  cp: cannot stat '-n': ...
cp -i /a.md /d        -> rc=1  cp: cannot stat '-i': ...
cp -P /a.md /d        -> rc=1  cp: cannot stat '-P': ...
cp -L /a.md /d        -> rc=1  cp: cannot stat '-L': ...
cp --parents /a.md /d -> rc=1  cp: cannot stat '--parents': ...
cp -t /d /a.md        -> rc=1  cp: target '/a.md': Not a directory
```

GNU exits 0 for `-v`/`-n`/`-i`/`-P`/`-L`/`-t` and prints `cp: invalid option -- 'Q'` for a genuinely unknown one. This is the pre-existing `FlagParser` convention rather than something the PR broke (before it, `cp -v a b` was `missing file operand`), but since this PR is where `cp` learns flags, accepting `-v`/`-n`/`-i`/`-P`/`-L` as no-ops (or at least `-v`, which scripts and agents use a lot) would be a cheap win.

lib/just_bash/commands/cp.ex
```diff
@@ -1,31 +1,180 @@
...
+
+  defp copy_dir(bash, {src, src_path}, {dest, dest_path}, true) do
+    case FS.cp(bash.fs, src_path, dest_path, recursive: true) do
+      {:ok, fs} ->
+        {Command.ok(), %{bash | fs: fs}}
+
+      {:error, %VFS.Error{kind: :einval}} ->
+        {Command.error("cp: cannot copy a directory, '#{src}', into itself, '#{dest}'\n"), bash}
+
+      {:error, %VFS.Error{kind: :enotdir}} ->
```
davydog187, line 278: Note for the #56 merge (no change needed today). This wording is exactly GNU's for "destination exists as a regular file", and the fixture proves it. But once #56 makes `FS.Memory` return `:enotdir` when an *ancestor* component is a regular file, `cp_directory`'s `mkdir(parents: true)` will funnel that case here too, and the message will be wrong:

```
# with /f a regular file, post-#56
cp -r /s /f/sub   ->  cp: cannot overwrite non-directory '/f/sub' with directory '/s'
# GNU:                cp: cannot stat '/f/sub': Not a directory
```

Same shape applies to `file_result/3`'s fallback below (GNU: `cannot stat 'f/sub.md': Not a directory`, not `cannot create regular file ...`). Distinguishing "dest exists as a non-directory" from "a path component is not a directory" would need `lstat` on the dest before the copy — probably a follow-up, but worth a comment so the next person doesn't read the current wording as verified for both cases.

lib/just_bash/commands/cp.ex
```diff
@@ -1,31 +1,180 @@
...
+        fs
+        |> FS.cp(src_path, dest_path)
+        |> file_result(%{bash | fs: fs}, dest)
+
+      {:error, %VFS.Error{} = error} ->
+        {Command.error("cp: cannot stat '#{src}': #{FS.strerror(error)}\n"), bash}
+    end
+  end
+
+  defp copy_dir(bash, {src, _src_path}, {_dest, _dest_path}, false),
```
davydog187: Style/convention nit: `recursive?` is a positional boolean threaded through `copy/3`, `copy_each/4`, `copy_one/4`, `copy_resolved/4` and `copy_dir/4`, and these two clause heads dispatch on bare `false` / `true` — which is the exact pattern CLAUDE.md lists under "AVOID: boolean parameters" (`def fetch(id, true)`). Passing the flags map through, or `:recursive | :no_recursive`, would make the heads self-describing (`defp copy_dir(bash, src, dest, :no_recursive)`).

lib/just_bash/commands/cp.ex
```diff
@@ -1,31 +1,180 @@
...
+
+      {{:error, error}, bash} ->
+        {Command.error("cp: target '#{dest}': #{FS.strerror(error)}\n"), bash}
+
+      {_kind, bash} ->
+        {Command.error("cp: target '#{dest}': Not a directory\n"), bash}
     end
   end
+
+  defp dest_kind(bash, dest) do
```
davydog187, line 151: Minor shape nit: the first tuple element is either a kind atom (`:directory`, `:other`) or an error tuple, so callers have to pattern match two different kinds of thing in one slot — and the two-operand clause above silently absorbs the error case through `{_kind, bash}`. `{:ok, :directory | :other, bash} | {:error, error, bash}` would make that fall-through explicit. Behaviour is right either way (a missing destination correctly falls through to `copy_one`, which reports `cannot stat` for a missing source or creates the file).

lib/just_bash/fs/fs.ex
```diff
@@ -249,6 +260,11 @@ defmodule JustBash.FS do
 
   # ── private ──────────────────────────────────────────────────────────────
 
+  # Is `dest` the same path as `src`, or nested inside it? Everything is
+  # inside the root, so a recursive copy of "/" never has a safe destination.
+  defp within?(_dest, "/"), do: true
+  defp within?(dest, src), do: dest == src or String.starts_with?(dest, src <> "/")
```
davydog187, line 291: Verified this guard rather than trusting it, since a prefix test is easy to get wrong:

- No sibling-prefix false positive: `mv /ab /abc` and `cp -r /ab /abc` both still work (the `src <> "/"` suffix does its job) — nice that `fs_test.exs` pins this.
- Direction is right: `dest` inside `src` is the unsafe one, and the legal upward copy `cp -r /a/b/d /a` still succeeds (`/a/d/x.md`).
- `within?(_dest, "/") -> true` reads over-broad but is correct here: in a single-tree FS every destination is inside the root, so a recursive copy of `/` genuinely has no safe target. Divergence from GNU (which copies and skips the destination subtree) is fine; `cp -r / /backup` returning `cannot copy a directory, '/', into itself, '/backup'` is a reasonable answer.
- No hang: `cp -r /a /a/b/deep/x` (dest several levels down, not yet existing) is rejected, and a tree containing a symlink to itself copies fine because symlinks are copied as links.

davydog187 (commented):
## Review: `cp` into a directory + real `-r` (#52)

Checked out the branch, re-ran the gates, and probed the new behaviour against **GNU coreutils cp 9.x** (`/opt/homebrew/opt/coreutils/libexec/gnubin/cp`, so GNU semantics, not the BSD `cp` that macOS ships) plus the in-memory FS directly. Gates on my checkout: `mix compile --warnings-as-errors` clean, `mix format --check-formatted` clean, `mix credo --strict` clean except the pre-existing `test/support/banned_fixture_apply.ex` fixture finding, `mix test` = **3798 tests, 56 properties, 0 failures**. (Dialyzer not re-run.)

The two headline fixes are right and the error wording is impressively faithful. What I verified against real GNU `cp`, all matching this PR exactly:

| case | GNU | this PR |
|---|---|---|
| `cp -r a a` | `cannot copy a directory, 'a', into itself, 'a/a'` | same |
| `cp -r a a/b` | `cannot copy a directory, 'a', into itself, 'a/b/a'` | same |
| `cp -r a/b a` | `'a/b' and 'a/b' are the same file` | same |
| `cp ./f f` | `'./f' and 'f' are the same file` | same |
| `cp nope1 nope2 d` | both messages, rc=1, other sources still copied | same |
| `cp s d -r` (flags after operands) | works | works |
| `cp -- ./-r rr` | works | works |
| multi-source into a symlinked dir | works | works |

Also confirmed the subtree guard does **not** over-match: `cp -r /a/b/d /a` (legal upward copy) still succeeds, `mv /ab /abc` and `cp -r /ab /abc` are unaffected, and `cp -r /a /a/b/deep/x` plus a self-symlink inside a copied tree both terminate instead of hanging. `within?(_, "/")` looked over-broad at first glance but is exactly right for a single-tree FS — every destination really is inside the root.

Two things I'd fix before merge, both in the new symlink path; the rest are notes.

### Findings

**1. Medium — `cp -r` on a symlink-to-directory copies nothing** (`cp.ex:115-116`, `cp.ex:148-158`). The symlink branch is taken before any directory check and unconditionally `read_file`s the target, so a link to a directory dies with `Is a directory` and no copy happens. GNU copies it (as a link) with `-r`, and gives the standard omit message without.

**2. Medium/low — a dereferenced symlink copy drops the source mode** (`cp.ex:152`). `FS.write_file/3` with no `mode:`/`mtime:` lands 0644; GNU keeps 0755. This contradicts the new moduledoc claim that a copy "always carries the source's mode and mtime across".

**3. Low — an empty destination operand silently copies to the wrong place** (`cp.ex:171-176`). `cd /work && cp /x/a.md ''` exits **0** and writes `/a.md` (the root), leaving `/work` untouched; GNU exits 1 with `cannot create regular file '': No such file or directory`.

**4. Low — GNU options that now parse as filenames** (`cp.ex:22-33`): `-v`, `-n`, `-i`, `-P`, `-L`, `-d`, `-t`, `--parents` all end up as operands, so e.g. `cp -v a.md d` exits 1 with `cp: cannot stat '-v'...`. Pre-existing `FlagParser` convention (unknown flags fall through), but flag parsing is what this PR just introduced for `cp`, so it's the natural place to at least no-op `-v`/`-n`/`-i`/`-P`/`-L`.

**5. Conventions** — `recursive?` is a positional boolean threaded through five private functions, with `copy_dir(..., false)` / `copy_dir(..., true)` clause heads. CLAUDE.md calls out exactly this ("AVOID: boolean parameters"). Passing the flags map, or `:recursive | :no_recursive`, would read better at the `copy_dir` heads. Also `dest_kind/2` mixes kind atoms and an error tuple in the same tuple slot, and the two-operand clause silently absorbs the error case via `{_kind, bash}`.

**6. Fixture coverage** — the recorded suite is a real gain (there was none for `cp`), but several of the most interesting new strings are asserted only in unit tests, with no bash case pinning them: `cp -r a/b a` ("are the same file"), `cp -r a a`, `--recursive`, `-a`, `--`, the symlink dereference, mode-via-symlink, and the `mv` subdirectory guard. I checked the two "into itself"/"same file" wordings against GNU by hand and they match exactly — so the assertions are correct, they're just not locked by a fixture. Worth adding a few, especially the `mv` guard.

### Interaction with #56 (not a conflict, but a semantic note)

Textual conflict in `cp.ex` is trivial: #57 already has a catch-all `file_result/3` that produces #56's exact `cp: cannot create regular file '<dest>': <strerror>` message, so #56's `write_dest/4` just disappears. Semantically they're complementary — #56 is what makes #57's `:enotdir` branches reachable. Two wording notes for whoever resolves it:

- After #56, `cp -r /s /f/sub` (where `/f` is a regular file) will return `:enotdir` from `cp_directory`'s `mkdir(parents: true)` and hit `copy_dir`'s `:enotdir` clause → `cannot overwrite non-directory '/f/sub' with directory '/s'`. GNU says `cp: cannot stat '/f/sub': Not a directory`. That clause conflates "destination exists as a file" (where the current wording is right, and the fixture proves it) with "an ancestor component is a file". Same for `file_result`'s fallback: GNU says `cannot stat 'f/sub.md': Not a directory`, not `cannot create regular file ...`.
- One gap #56 won't close, because `resolve_path/2` strips the trailing slash: `cp /a.md /f/` where `/f` is a regular file currently **overwrites `/f`** and exits 0. GNU: `cp: cannot stat 'f/': Not a directory`. POSIX says a trailing slash requires a directory. Worth a follow-up issue rather than growing this PR.

lib/just_bash/commands/cp.ex
davydog187: Fixed in 8396103. You were right on both halves of this, including that the code comment had the rule backwards.

Link handling is now a resolved decision rather than a branch order accident — `%Cp.Options{links: :follow | :preserve}`, set from `-L` > `-P`/`-d` > (`-r`/`-a` ⇒ `:preserve`) > shallow ⇒ `:follow`:

```elixir
# A preserved link is copied as a link — dangling or not, directory or not.
defp copy_link(bash, {src, src_path}, {dest, dest_path}, %Options{links: :preserve} = opts) do
  bash.fs |> FS.cp(src_path, dest_path) |> file_result(...)
end

# A followed link copies whatever it names, under the link's own name: a link
# to a directory therefore needs `-r`, exactly like a directory does.
defp copy_link(bash, {src, src_path}, dest, %Options{links: :follow} = opts) do
  case link_target(bash.fs, src_path, 0) do
    {:ok, target_path, fs} -> copy_resolved(%{bash | fs: fs}, {src, target_path}, dest, opts)
    ...
```

Re-entering `copy_resolved/4` with the resolved target but the link's own spelling means the omit message falls out for free and stays in the operand's words. Behaviour now, all pinned by recorded fixtures:

| case | bash | now |
|---|---|---|
| `cp sdlink outdir` | `-r not specified; omitting directory 'sdlink'` | same |
| `cp -r sdlink outdir` | rc=0, `outdir -> sd` | same |
| `cp -r alink rcopy` / `cp -a alink acopy` | link | same |
| `cp -P alink pcopy` / `cp -d alink dcopy` | link | same |
| `cp -rL sdlink ldir` | real directory | same |
| `cp -r dangling dcopy` | rc=0, link | same |
| `cp alink acopy` | dereferenced, mode kept | same |

`link_target/3` walks the chain (bounded at 32 hops → `:eloop`) rather than a single `readlink`, so a chain of links resolves and a cycle reports instead of spinning.

One thing worth knowing for the fixture: I could not assert reading *through* a copied directory symlink, because the FS does not resolve symlinks in intermediate path components at all — `cat /sdlink/x.md` fails on the original link too. So those cases assert `readlink` and `[ -L ]`, and the fixture scripts avoid `ls outdir/`.

lib/just_bash/commands/cp.ex
davydog187: Fixed in 8396103, exactly the way you suggested — resolving the link and delegating to `FS.cp/4` on the target, so this path picks up `cp_regular`'s `mode:`/`mtime:` instead of a bare `write_file/3`. Findings 1 and 2 really were one bug.

New test:

```elixir
test "cp through a symlink keeps the target's mode" do
  # /src.sh written with mode: 0o755, /mlink -> /src.sh
  {result, bash} = JustBash.exec(bash, "cp /mlink /mcopy")
  assert result.exit_code == 0
  {:ok, stat, _fs} = FS.stat(bash.fs, "/mcopy")
  assert stat.mode == 0o755
end
```

Also pinned by a recorded fixture (`cp alink acopy; test -x acopy && echo exec; readlink acopy`), so the moduledoc's "always carries the source's mode and mtime across" is now true on every path rather than every path but one.

lib/just_bash/commands/cp.ex
davydog187: Fixed in 8396103. Took the "reject an empty operand up front" route, since the spelling is what every message has to quote anyway:

```elixir
defp copy_one(bash, "", _dest, _opts),
  do: {Command.error("cp: cannot stat '': No such file or directory\n"), bash}

defp copy_one(bash, src, "", opts), do: empty_dest(bash, src, opts)

# An empty operand names no file. Resolving it would produce the cwd, which is
# not what bash does with it.
defp dest_kind(bash, ""), do: {:error, VFS.Error.new(:enoent, path: ""), bash}
```

Matching bash, all three recorded as fixtures:

- `cp /x/a.md ''` → rc=1, `cp: cannot create regular file '': No such file or directory` (and `/a.md` is not created)
- `cp '' /d` → rc=1, `cp: cannot stat '': No such file or directory`
- `cp /a.md /b.md ''` → rc=1, `cp: target '': No such file or directory`

One deliberate divergence, marked with a comment in the test: `cp -r s ''` exits **0** in GNU 9.x and copies nothing, which looks more like a quirk than an intent, so this returns rc=1 with `cannot create directory ''` instead. No fixture for that one, precisely because I did not want to lock in bash's answer there.

lib/just_bash/commands/cp.ex
davydog187: Addressed in 8396103 — agreed that this PR is the right place, since it is where `cp` learns flags at all. Rather than no-op them I implemented the ones whose semantics are cheap, because a silent no-op for `-n` would *lose data* (overwrite where bash would not) and a silent `-v` would drop output a caller asked for:

- `-v`/`--verbose` — real output: `'src' -> 'dest'` per copied entry, parents before children, walking the source the way bash does. The recorded fixtures pin all three shapes, including the nested `cp -rv s d` (`'s' -> 'd/s'`, `'s/n' -> 'd/s/n'`, `'s/n/n.md' -> ...`, `'s/y.md' -> ...`) — the memory backend's `readdir` sorts, so the order matches GNU's.
- `-n`/`--no-clobber` — really skips an existing destination. Coarser than bash: decided per operand, not per file inside a tree. Documented in the moduledoc; the fixtures use `ignore_stderr` because GNU 9.x prints a `-n is non-portable` warning we do not.
- `-P`/`--no-dereference` (and `-d`) / `-L`/`--dereference` — real link handling, see the thread on finding 1.
- `-i`/`--interactive` — accepted as a no-op that proceeds. This is the one honest divergence: with no stdin GNU answers the prompt "no" and exits 1, and a sandbox has no terminal to prompt on. Called out in the moduledoc and in a test comment; no fixture, since bash's answer here depends on the tty.

Left unparsed on purpose: `-t`/`--target-directory` and `--parents`. Both change what the operands *mean* (no trailing destination; destination layout mirrors the source path), so they want their own tests and fixtures rather than a line in the flag spec.

Also left alone: `cp -Q` still reports `cannot stat '-Q'` instead of `invalid option -- 'Q'`. Telling an unknown flag from a file named `-Q` after `--` needs a pre-scan that `FlagParser` does not expose, and every command in the tree shares the fall-through convention — worth fixing in `FlagParser` for all of them rather than here.

lib/just_bash/commands/cp.ex
davydog187: Recorded as a comment in 8396103 rather than a behaviour change, since as you say telling the two apart needs an lstat of the destination first:

```elixir
# Verified against bash for a destination that already exists as a regular
# file. A destination whose *ancestor* is a regular file surfaces the same
# kind, and bash words that one "cannot stat '<dest>': Not a directory" —
# telling the two apart needs an lstat of the destination first, so read
# this wording as unverified for that case.
{:error, %VFS.Error{kind: :enotdir}} ->
```

…and a shorter pointer to it above `file_result/3`'s fallback, which has the same shape.

That keeps the wording that the fixture proves, and stops the next reader from taking it as verified for the ancestor case. Whoever lands #56 gets the note at the exact line they will be looking at; if you want it as a tracked item rather than a comment I am happy to open one.

lib/just_bash/commands/cp.ex
davydog187: Fixed in 8396103. The positional boolean is gone; a struct carries the resolved flags, so the clause heads say what they mean:

```elixir
defmodule Options do
  @moduledoc false
  @enforce_keys [:mode, :links, :clobber, :verbose?]
  defstruct [:mode, :links, :clobber, :verbose?]
  # mode: :recursive | :shallow, links: :follow | :preserve,
  # clobber: :always | :never, verbose?: boolean()
end

defp copy_dir(bash, {src, _src_path}, {_dest, _dest_path}, %Options{mode: :shallow}),
  do: {Command.error("cp: -r not specified; omitting directory '#{src}'\n"), bash}

defp copy_dir(bash, {src, src_path}, {dest, dest_path}, %Options{mode: :recursive} = opts) do
```

It paid for itself immediately — the new flags (`-n` → `clobber`, `-P`/`-L`/`-d` → `links`, `-v` → `verbose?`) went in as struct fields with dispatching heads (`%Options{clobber: :never}`, `%Options{links: :preserve}`) instead of four more positional booleans, which is the shape the anti-pattern list is warning about. Nested `Options` module with `@moduledoc false` follows `ArgParser.Context`'s precedent; the registry maps names explicitly so nothing picks it up as a command.

lib/just_bash/commands/cp.ex
davydog187: Fixed in 8396103, using the shape you proposed:

```elixir
defp dest_kind(bash, dest) do
  case FS.stat(bash.fs, FS.resolve_path(bash.cwd, dest)) do
    {:ok, %VFS.Stat{type: :directory}, fs} -> {:ok, :directory, %{bash | fs: fs}}
    {:ok, %VFS.Stat{}, fs} -> {:ok, :other, %{bash | fs: fs}}
    {:error, %VFS.Error{} = error} -> {:error, error, bash}
  end
end
```

Both callers now name every case, so the two-operand fall-through is deliberate on its face rather than a wildcard:

```elixir
{:ok, :directory, bash} -> copy_one(bash, src, dest_in_dir(dest, src), opts)
{:ok, :other, bash} -> copy_one(bash, src, dest, opts)
{:error, _error, bash} -> copy_one(bash, src, dest, opts)
```

The empty-operand guard from finding 3 landed as an extra `dest_kind(bash, "")` head, which the old shape would have made harder to read.

lib/just_bash/fs/fs.ex
davydog187: Thanks for probing it rather than trusting it — the sibling-prefix case (`/ab` vs `/abc`) is exactly why the `src <> "/"` suffix is in there, and the upward copy `cp -r /a/b/d /a` is the case a naive `String.contains?` would have broken.

Nothing changed here in 8396103, and your read of `within?(_dest, "/")` matches the comment's intent. Worth adding: enabling `-r` also made the guard load-bearing for a case neither of us listed — a *symlink* to an ancestor inside the tree. That one terminates for the reason you spotted (links are copied as links, never followed during the walk), and the new `link_target/3` for `-L`/shallow copies is bounded at 32 hops so a link cycle reports `:eloop` instead of spinning.

davydog187:
Thanks for the review — the GNU 9.x probing caught two real bugs in the new symlink path, and one of your fixture suggestions caught a third that neither of us had listed. All addressed in 8396103; replies are on each thread, summary here.

| # | finding | outcome |
|---|---|---|
| 1 | `cp -r` on a symlink-to-directory copies nothing; wrong shallow message | **Fixed.** Link handling is now a resolved `links: :follow \| :preserve` decision instead of a branch order; preserved links delegate to `FS.cp/4`, followed links resolve the chain and copy what it names, so a link to a directory needs `-r` like a directory does. |
| 2 | dereferenced symlink copy drops the source mode | **Fixed** by the same change — the followed path goes through `FS.cp/4`, so `mode:`/`mtime:` come across. |
| 3 | empty destination operand lands in the wrong directory, exit 0 | **Fixed.** An empty operand names no file: `cannot stat ''` / `cannot create regular file ''` / `target '': No such file or directory`. |
| 4 | GNU options parse as filenames | **Fixed for `-v`, `-n`, `-i`, `-P`, `-L`, `-d`** — implemented rather than no-oped where semantics allowed (a no-op `-n` would overwrite where bash would not). `-t` and `--parents` **deferred**: both change what the operands mean. `invalid option -- 'Q'` **deferred** to `FlagParser`, which all commands share. |
| 5 | positional `recursive?` boolean; `dest_kind/2` shape | **Fixed.** `%Cp.Options{mode, links, clobber, verbose?}` struct with dispatching clause heads; `dest_kind/2` returns `{:ok, kind, bash} \| {:error, error, bash}` and both callers name every case. |
| 6 | fixture gaps | **Fixed.** 23 more recorded `cp` cases + the `mv` guard in `mv.json` (43 `cp` + 7 `mv` total, all green). |
| — | #56 `:enotdir` wording note | **Comment**, not a behaviour change — distinguishing "destination is a file" from "ancestor is a file" needs an lstat of the destination first. |
| — | trailing slash onto a regular file | **Follow-up issue #58** — pre-existing `resolve_path/2` behaviour, and `mv`/redirections share it. |

Recording the `mv` guard fixture you asked for immediately earned its keep: the message printed the *resolved* destination where bash prints the operand spelling.

```
expected: mv: cannot move 'a' to a subdirectory of itself, 'a/b/a'
actual:   mv: cannot move 'a' to a subdirectory of itself, '/tmp/jb_mv_3330/a/b/a'
```

`Mv` now tracks `dest_final` (where the move lands) and `dest_shown` (the same place in the operand's words) separately. That bug was invisible to unit tests written with absolute paths — the fixture was the only thing that could have caught it.

Two divergences deliberately left in, both commented in the code and in the tests, and neither fixtured so bash's answer is not locked in:

- `cp -r s ''` — GNU 9.x exits **0** and copies nothing; this exits 1 with `cannot create directory ''`.
- `cp -i` when the destination exists — GNU with no stdin answers the prompt "no" and exits 1; a sandbox has no terminal, so `-i` proceeds as if answered yes.

Gates on the updated branch: `mix compile --warnings-as-errors` clean, `mix format --check-formatted` clean, `mix credo --strict` clean except the pre-existing `banned_fixture_apply.ex` fixture finding, `mix dialyzer` 17 errors / 17 skipped (identical to `main`), `mix test` **3841 tests, 56 properties, 0 failures**, and green again with `--include bash_comparison`.

ivarvong:
Two things once #56 lands. I rebased locally to check — the conflicts are `cp.ex` and `mv.ex`, and taking #57's `cp.ex` wholesale plus keeping both `mv` error clauses resolves them. On that resolution the merged suite is 3,932 tests / 1 failure, and one shape hangs that no test covers.

**1. `within?/2` is a lexical prefix test, so a symlinked destination walks around it.**

```elixir
b = JustBash.new(cwd: "/w")
JustBash.exec(b, "mkdir -p /w/a; ln -s /w/a /w/l; cp -r /w/a /w/l/x")
# no result in 90s — each pass creates children under the tree it is still walking
```

Same for `cp -r /w/a /w/alink/sub`, `cp -r /w/a /w/alink`, and `mv /w/a /w/alink/sub` with `alink -> /w/a`. GNU 9.10 exits 1 on all of them: `cp: cannot copy a directory, 'a', into itself, 'alink/sub/a'`, `mv: cannot move 'a' to a subdirectory of itself, 'alink/sub/a'`.

On main's FS today those same scripts exit 0 and quietly copy nothing — except `mv`, which destroys the source:

```elixir
JustBash.exec(JustBash.new(cwd: "/w"), "mkdir -p /w/a; echo hi > /w/a/f; ln -s /w/a /w/l; mv /w/a /w/l")
# exit 0; afterwards FS.read_file("/w/a/f") and FS.read_file("/w/l/f") both error, and ls /w shows only l.
# GNU exits 1 and leaves a/f intact.
```

`fs.ex:265` compares normalized strings, so `/w/l/x` never looks like it is inside `/w/a`. #56's component-wise resolution is what makes the real fix available: resolve both operands before the prefix test at `fs.ex:196`.

**2. The merged suite's one failure — `not_a_directory_test.exs:283`:**

```
left:  "cp: cannot create regular file '/m/j/a.md': Not a directory"
right: "cp: cannot stat '/m/j/a.md': Not a directory"
```

GNU stats the destination first, so an ancestor that is a regular file surfaces as `cannot stat`. That is the `file_result` catch-all at `cp.ex:287` wanting an `:enotdir` clause, and `copy_dir` at `cp.ex:270` wants the same split its own comment predicts.

For what it's worth, the 42 cp/mv scenarios I ran against GNU 9.10 otherwise matched — dir merges, `src/.`, every `-r`/`-a` form, `-P`/`-L` link rules, dangling links both ways, multi-source partial failure, `-v` over a whole tree. Two small ones beyond what you have already disclosed or filed as #58, which I will drop in as issues rather than hold this up: same-file detection is path equality, so `cp f l` with `l -> f` copies where GNU refuses, and an unknown flag becomes an operand (`cp -z f g` reports `target 'g': No such file or directory` rather than `invalid option -- 'z'`). And the pairing fixes the other direction too — `cp f <symlink-to-dir>` and `cp -r a <file>/sub` write silently unreachable entries at exit 0 on main's FS, and are correct once #56 is underneath.

ivarvong (approved):

PR #84, opened 2026-08-18
fix: grep BRE alternation, missing-path exit code, and flag-level did-you-mean

davydog187:
## What broke

A production agent ran `grep -ril "bodyweight\|weight\|weigh" /memory`, got
an empty result at exit 1, and concluded the fact wasn't recorded — it was,
in two of the files under `/memory`. Full writeup, including a second
downstream fix for the flag-level recovery path, is in
[the downstream issue](https://github.com/davydog187/dol/issues/80).

## 1. `grep`'s `\|` is BRE alternation, not an escaped literal pipe

`compile_pattern/3` (`lib/just_bash/commands/grep.ex`) passed the pattern
straight to `Regex.compile/2` (PCRE), where `\|` is an escaped literal pipe.
Real `grep` without `-E`/`-P` is BRE, where `\|` is alternation. The PCRE
pattern compiles cleanly either way, so the existing `{:error, _} ->
Regex.escape(...)` fallback never fires — the pattern just silently matches
a literal pipe character that occurs nowhere:

```elixir
Regex.compile!(~S(bodyweight\|weight\|weigh), "i")
|> Regex.match?("- **Bodyweight:** 208 lb, logged as a metric row 2026-08-18.")
#=> false          # before: silent miss, exit 1

Regex.compile!("bodyweight|weight|weigh", "i") |> Regex.match?(same)
#=> true           # what real BSD/GNU grep -ril returns on the same file
```

**Fix shipped:** rewrite only `\|` -> `|` when neither `-E` nor `-P` is
given. `\|` is the dominant agent idiom for alternation (and the one that
caused the incident above); full BRE emulation (bare `(`, `)`, `{` as
literals, `\(...\)` as groups, `\{n,m\}` as bounds, etc.) is a much larger
surface with more ways to land partially right, so it's deliberately left
alone rather than half-translated.

Also fixed on the same path: a missing or unreadable file was silently
swallowed (`{:error, _} -> {acc, had_match, fs}`), producing exit 1 with no
stderr — indistinguishable from "no match" from the caller's side. Real
grep exits 2 and names the path (`grep: FILE: No such file or directory`).
Fixed to match, including `-q`'s documented "exit 0 on a match even if an
error was also detected on another file" priority inversion.

Fixture-corpus cases recorded against real GNU grep (Ubuntu 24.04, in the
Docker corpus) cover both: alternation matching/non-matching, `-E`
opting out of the rewrite, and the missing-path exit code/stderr message.

## 2. An unknown flag now names the sibling command that declares it

Separately: `dol log metric --weight 207.8` (a plausible guess — the right
command is `dol log weight`) failed loudly with `unknown option: --weight`
and a usage line, but nothing pointed at the sibling leaf that actually
declares `--weight`. `allow_unknown_flags` already defaults to `false`, so
there's no silent-drop bug here — just a missing recovery pointer at the
single point (stderr) an agent reads to self-correct.

`ArgParser.parse/3` gains an `:on_unknown_flag` hook, called with the bare
flag name whenever it would otherwise be a hard error; a non-nil return is
appended to the message. `JustBash.CLI` wires it up in `dispatch_leaf/6`,
scanning sibling commands in the same group for one that declares the
flag — an exact declaration match, not the Jaro-distance fuzzy match
`Help.unknown_subcommand/4` already uses for subcommand-name typos (that's
the right tool for a name typo, not for "which leaf owns this flag").

```
dol log metric: unknown option: --weight — did you mean 'dol log weight'?
```

New tests in `test/cli/routing_test.exs` cover: a sibling's long flag, a
sibling's short flag, no suggestion when no sibling declares the flag, and
no cross-group leakage (a flag declared by a leaf in an unrelated group is
not suggested).

## Testing

- `mix test` — full suite green, 5086 passed / 5 excluded (pre-existing
  `:live` exclusions, unrelated to this change)
- `mix bash_fixtures.verify` — corpus sound
- `mix format --check-formatted` — clean on all changed files

### DIFF
--- a/lib/just_bash/cli.ex
+++ b/lib/just_bash/cli.ex
@@ -685,7 +685,7 @@ defmodule JustBash.CLI do
   defp dispatch_leaf(cli, %Command{} = leaf, path, rest, bash, stdin) do
     label = command_label(cli, path)
 
-    with {:ok, flags, positional, extra} <- parse_leaf_args(leaf, rest, label),
+    with {:ok, flags, positional, extra} <- parse_leaf_args(cli, leaf, path, rest, label),
          :ok <- validate_positionals(leaf.args, positional, label),
          invocation = build_invocation(leaf, flags, positional, extra, bash, stdin, path),
          :ok <- run_validate(leaf, invocation) do
@@ -699,15 +699,59 @@ defmodule JustBash.CLI do
   # Parse a leaf's flags, normalizing to a 4-tuple so the caller is uniform. A leaf that
   # opts into `allow_unknown_flags` collects undeclared flags into `extra`; otherwise `extra`
   # is always empty.
-  defp parse_leaf_args(%Command{allow_unknown_flags: true} = leaf, rest, label) do
-    ArgParser.parse(rest, leaf.flags, command: label, collect_unknown: true)
+  defp parse_leaf_args(cli, %Command{allow_unknown_flags: true} = leaf, path, rest, label) do
+    ArgParser.parse(rest, leaf.flags,
+      command: label,
+      collect_unknown: true,
+      on_unknown_flag: &sibling_flag_hint(cli, path, &1)
+    )
   end
 
-  defp parse_leaf_args(%Command{} = leaf, rest, label) do
-    case ArgParser.parse(rest, leaf.flags, command: label) do
+  defp parse_leaf_args(cli, %Command{} = leaf, path, rest, label) do
+    opts = [command: label, on_unknown_flag: &sibling_flag_hint(cli, path, &1)]
+
+    case ArgParser.parse(rest, leaf.flags, opts) do
       {:ok, flags, positional} -> {:ok, flags, positional, []}
       {:error, _} = err -> err
     end
+  end
+
+  # A leaf rejecting an undeclared flag is one hop from the fix when a sibling command in
+  # the same group declares that exact flag — e.g. `dol log metric --weight` when `dol log
+  # weight` is the leaf that takes `--weight`. This is not fuzzy matching like
+  # `Help.unknown_subcommand/4`'s Jaro suggestion: the flag either is or isn't declared
+  # elsewhere in the group, so an exact declaration match is the whole signal.
+  #
+  # Lives here (rather than in `ArgParser`) because only the CLI tree has a notion of
+  # "sibling commands in the same group" — `ArgParser.parse/3` only ever sees one leaf's
+  # flag spec, threaded in through the `:on_unknown_flag` hook.
+  defp sibling_flag_hint(cli, path, flag) do
+    current_name = List.last(path)
+    group_path = Enum.drop(path, -1)
+
+    cli.commands
+    |> resolve_group_commands(group_path)
+    |> Enum.reject(&(&1.name == current_name or Command.group?(&1)))
+    |> Enum.find(&flag_declared?(&1.flags, flag))
+    |> case do
+      nil -> nil
+      %Command{name: name} -> "did you mean '#{command_label(cli, group_path ++ [name])}'?"
+    end
+  end
+
+  defp resolve_group_commands(commands, []), do: commands
+
+  defp resolve_group_commands(commands, [name | rest]) do
+    case Enum.find(commands, &(&1.name == name)) do
+      %Command{commands: children} -> resolve_group_commands(children, rest)
+      nil -> []
+    end
+  end
+
+  defp flag_declared?(flags, flag) do
+    Enum.any?(flags, fn {_name, spec} ->
+      flag == spec[:short] or flag == spec[:long] or flag in (spec[:aliases] || [])
+    end)
   end
 
   defp build_invocation(_leaf, flags, positional, extra, bash, stdin, path) do
--- a/lib/just_bash/cli.ex
+++ b/lib/just_bash/cli.ex
@@ -720,7 +720,11 @@ defmodule JustBash.CLI do
   # the same group declares that exact flag — e.g. `dol log metric --weight` when `dol log
   # weight` is the leaf that takes `--weight`. This is not fuzzy matching like
   # `Help.unknown_subcommand/4`'s Jaro suggestion: the flag either is or isn't declared
-  # elsewhere in the group, so an exact declaration match is the whole signal.
+  # elsewhere in the group, so an exact declaration match is the whole signal — as long as
+  # that match is unique. `--verbose`, `--json` and `--force` are routinely declared by
+  # several leaves in one group, and there the first sibling in declaration order carries
+  # no more information than any other; a confidently-worded pointer at an arbitrary one
+  # is worse than none, so an ambiguous match yields no hint.
   #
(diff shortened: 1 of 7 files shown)


lib/just_bash/commands/grep.ex
```diff
@@ -230,19 +251,18 @@ defmodule JustBash.Commands.Grep do
...
-
-        flags.w ->
-          "\\b" <> pattern <> "\\b"
-
-        flags.x ->
-          "^" <> pattern <> "$"
-
-        true ->
-          pattern
+        flags.w -> "\\b" <> base_pattern <> "\\b"
```
davydog187: **`-w`/`-x` wrap an ungrouped alternation, so the anchor binds to only one branch.**

Now that `\|` compiles to a real `|`, `grep -w 'foo\|bar'` builds `\bfoo|bar\b`, which PCRE reads as `(\bfoo)|(bar\b)`.

| | this branch | GNU grep 3.11 |
|---|---|---|
| `echo foobar \| grep -w 'foo\|bar'` | matches, exit 0 | no match, exit 1 |
| `echo foobaz \| grep -x 'foo\|bar'` | matches, exit 0 | no match, exit 1 |

This is a false-positive class the PR creates: before the change `\|` was a literal, so `-w`/`-x` combined with `\|` could not produce a spurious match.

```elixir
flags.w -> "\\b(?:" <> base_pattern <> ")\\b"
flags.x -> "^(?:" <> base_pattern <> ")$"
```

lib/just_bash/commands/grep.ex
```diff
@@ -251,6 +271,27 @@ defmodule JustBash.Commands.Grep do
...
+  # This rewrites only `\|` -> `|`. Full BRE emulation — bare `(`, `)`, `{`
+  # as literals, `\(...\)` as groups, `\{n,m\}` as bounds — is a much larger
+  # surface with more ways to get it partially right, and is left alone
+  # rather than half-translated. `\|` is the dominant idiom and the one that
+  # caused the incident this fixes, so it is handled on its own.
+  defp bre_alternation_to_pcre(pattern, flags) do
+    if flags.e_ext or flags.p_pcre do
+      pattern
+    else
+      String.replace(pattern, "\\|", "|")
```
davydog187: **The substring replace also consumes the pipe after an escaped backslash.**

`a\\|b` in BRE is a literal backslash followed by a literal `|`. The blind `String.replace/3` rewrites it to `a\|b`, which PCRE reads as the literal string `a|b`:

```
$ echo 'a|b' | grep 'a\\|b'
# this branch: matches, exit 0
# GNU grep:    no match, exit 1
```

A scan that skips the character after each backslash — or `Regex.replace(~r/\\(.)/, ...)` with a replacement function — handles it.

Related, and pre-existing rather than introduced here: bare `|` is still passed straight through as alternation, so after this PR there is no spelling that matches a literal pipe in default BRE mode. Worth escaping bare `|` in the same pass.

lib/just_bash/commands/grep.ex
```diff
@@ -172,13 +173,22 @@ defmodule JustBash.Commands.Grep do
...
   end
 
+  # A missing or unreadable path is an error real grep reports on stderr at
+  # exit 2, not a silent "no match" at exit 1 — an agent cannot tell those
+  # apart from `grep`'s output alone otherwise.
+  defp format_file_error(file, %VFS.Error{kind: :eisdir}),
+    do: "grep: #{file}: Is a directory\n"
+
+  defp format_file_error(file, _error),
+    do: "grep: #{file}: No such file or directory\n"
```
davydog187: **Every non-`:eisdir` error is reported as "No such file or directory"**, which reintroduces exactly the bug `FS.strerror/1` exists to prevent.

`FS.read_file/2` can return `:enotdir`, `:eacces`, `:eloop`, `:erofs`, and `:eio`, and `FS.strerror/1` (`lib/just_bash/fs/fs.ex:345-358`) already accepts a `%VFS.Error{}` directly. Verified with `grep hi /a.txt/x` where `/a.txt` is a regular file:

```
this branch: grep: /a.txt/x: No such file or directory
GNU grep:    grep: /a.txt/x: Not a directory
```

A `chmod 000` file likewise gives GNU `Permission denied`. Both clauses collapse to:

```elixir
defp format_file_error(file, error), do: "grep: #{file}: #{FS.strerror(error)}\n"
```

lib/just_bash/commands/grep.ex
```diff
@@ -189,15 +199,26 @@ defmodule JustBash.Commands.Grep do
...
-    exit_code = if any_match, do: 0, else: 1
+  defp build_files_result(bash, results, any_match, errors, flags) do
+    stderr = errors |> Enum.reverse() |> Enum.join()
+
+    # -q exits 0 on the first match found even if an error was also
+    # detected among the other files (documented GNU grep behavior);
+    # otherwise a file-level error outranks "no match found".
+    exit_code =
+      cond do
+        flags.q and any_match -> 0
```
davydog187: **`-q` reports errors for files real grep never opens.**

The contract is "exit *immediately* with zero status if any match is found", so GNU stops at the first match and never stats the later operands. With `a.txt` matching:

```
$ grep -q hi a.txt nope.txt
# GNU:         exit 0, empty stderr
# this branch: exit 0, but "grep: nope.txt: No such file or directory" on stderr
```

The exit code is right, the diagnostic is spurious. Suppressing `stderr` when `flags.q and any_match` is the cheap approximation of the short-circuit.

lib/just_bash/cli.ex
```diff
@@ -699,17 +699,61 @@ defmodule JustBash.CLI do
...
+  # "sibling commands in the same group" — `ArgParser.parse/3` only ever sees one leaf's
+  # flag spec, threaded in through the `:on_unknown_flag` hook.
+  defp sibling_flag_hint(cli, path, flag) do
+    current_name = List.last(path)
+    group_path = Enum.drop(path, -1)
+
+    cli.commands
+    |> resolve_group_commands(group_path)
+    |> Enum.reject(&(&1.name == current_name or Command.group?(&1)))
+    |> Enum.find(&flag_declared?(&1.flags, flag))
```
davydog187: **`Enum.find` names an arbitrary sibling when several declare the same flag.**

`--verbose`, `--json`, and `--force` are commonly declared by many leaves in one group. The lookup returns the first sibling in declaration order, so `dol log metric --verbose` yields a confidently-worded `did you mean 'dol log <whichever-is-first>'?` that is very likely not the command the caller wanted.

The comment just above makes the case that an exact declaration match is an unambiguous signal — that justification only holds when the match is unique. Suggest `Enum.filter` and suppressing the hint (or listing all candidates) when more than one sibling matches.

davydog187 (commented):
Reviewed against real GNU grep 3.11 (`ubuntu:24.04`). The branch's full suite (5022 tests) and `mix bash_fixtures.verify` both pass, so everything below is a gap the added tests don't cover.

The one that matters is the `-w`/`-x` interaction: enabling `\|` as real alternation means those flags now wrap an ungrouped pattern, which is a false-positive class this PR itself introduces. The rest are smaller divergences from GNU.

